PR #4, opened 2019-09-26
[12.0][FIX] storage_file: Fix backend view (alert role)

rousseldenis:
This solves : Invalid XML: An alert (class alert-*) must have an alert, alertdialog or status role.


rousseldenis:
@simahawk

lmignon (approved):
The warning is no more there into the logs.

astirpe (approved):

PR #12, opened 2019-10-01
[12.0] add pylint-odoo pre-commit hook

sbidoul:
Add the default pylint config in `.pylintrc`.

The benefit of having it in the repo is that IDE can use it without any manual configuration.

still to be investigated
- [x] how to handle "beta" pylint-odoo messages from MQT


.travis.yml
```diff
@@ -1,52 +1,31 @@
...
-    - stage: linting
-      name: "Pre-Commit"
-      python: 3.6
-      install: pip install pre_commit
-      script: pre-commit run --all
 
 env:
   global:
-  - VERSION="12.0" TESTS="1" LINT_CHECK="0"
+  - VERSION="13.0" TESTS="0" LINT_CHECK="0" MAKEPOT="0"
```
lmignon: Why verson 13?

lmignon (commented):
@sbidoul Thank you. that looks great.Why do you try to test with odoo 13?

.travis.yml
sbidoul: Silly mistake, fixed.

sbidoul:
For beta pylint message my current idea is this:

1. have `.pylintrc` with all messages enabled, so an IDE that picks it will show all warnings
2. have `.pylintrc-mandatory` also present, with only mandatory checks
3. have pre-commit run pylint with `.pylintrc-mandatory`
4. have pre-commit run pylint a second time with `.pylintrc` and `--exit-zero', so pre-commit (and thus travis) will still show all beta messages

sbidoul:
Let's wait until the config is stabilized, then don't forget to squash. It's important to have all that in one commit.

lmignon:
@sbidoul Yes, I'm waiting. My changes are to fix the tests...

.travis.yml
```diff
@@ -28,8 +28,8 @@ jobs:
       python: "3.6"
       install: pip install pre-commit
       script: pre-commit run --all --show-diff-on-failure
-      after_success: true
-      before_install: true
+      after_success:
+      before_install:
```
sbidoul, line 32: @lmignon FYI. This should clarify that after_success and before_install are overrides of the the global default, and not a boolean flag (it was actually running the true command).

simahawk:
@sbidoul can you merge this? :pray:

sbidoul:
Yes of course. BTW, Only @sebastienbeau is part of this PSC. We should invite others.

/ocabot merge

PR #13, opened 2019-10-14
[WIP][MIG] storage_backend_s3: Migration to 12.0

simahawk:
Tests failing because it tries to call https://ak-testing-bucket.s3.eu-west-3.amazonaws.com w/out proper rights.
@sebastienbeau  can you help here?

### DIFF
--- a/storage_backend_s3/__init__.py
+++ b/storage_backend_s3/__init__.py
@@ -1,4 +1,2 @@
-# -*- coding: utf-8 -*-
-
 from . import models
 from . import components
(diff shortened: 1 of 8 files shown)


sebastienbeau:
I will take a look at it tomorrow
I see two possibility
- finding a way to share an IAM user of amazon 
- using a fake S3 server like this : https://github.com/jubos/fake-s3

simahawk:
as I've reported on https://github.com/OCA/storage/pull/14 tests pass without VCRpy. Calling s3 buckets directly works: we might want to get rid of VCRpy.

simahawk:
I've fixed tests here https://github.com/OCA/storage/pull/14 together w/ some refactoring and improvements.

PR #15, opened 2019-10-15
[12.0][FIX] storage_image_relation: Fix domain into product_image_relation_…

lmignon:
…form to avoid js erreor

ping @simahawk


simahawk (approved):
tested, works, thanks :)

PR #16, opened 2019-10-16
 storage_backend_sftp: refactor tests w/ mock

simahawk:
Includes fix for #7
ping @lmignon @gurneyalex

PR #17, opened 2019-10-21
[10.0] sftp tests + OCA's pre-commit

lmignon:
backport of #16 
+ uses pre-commit  config defined for OCA


lmignon:
Fast tracking to make runbot :green_apple:

PR #21, opened 2019-10-21
[13.0][MIG] storage_backend 

Laurent-Corron:
- [x] requires [component](https://github.com/OCA/connector/pull/355)

PR #14, opened 2019-10-14
[MIG] storage_backend_s3: Migration to 12.0 + improvements

simahawk:
Includes and superseeds #13

PR #26, opened 2019-10-23
[13.0][MIG] storage_backend_s3 + fake s3

mileo:


simahawk:
@mileo please have a look here https://github.com/OCA/storage/pull/14 where I've already added support for custom endpoint (+ other goodies).

As I mentioned in my last comment tests pass smoothly when running against a real bucket. They fail when using VCRpy. So, unless someone finds a silver bullet to make them work with it, I think we can simply mock them ([`moto`](https://github.com/spulec/moto) lib might help)

simahawk (changes requested):
test problem solved and code improved here https://github.com/OCA/storage/pull/14.
Can you rebase on 12.0 as it has been merged?

lmignon:
@mileo I'll finalize your PR once #21 and  #27 will be merged...

mileo:
Hi @lmignon and @simahawk 

I rebase this PR with 13.0 and make module installable again and make some improvements to use fake s3 that need the region name.

Like this example: https://blog.stigok.com/2019/06/07/using-scaleway-object-storage-s3-with-python3-boto3.html

storage_backend_s3/views/backend_storage_view.xml
```diff
@@ -6,12 +6,20 @@
...
-                <field name="aws_cache_control"/>
+                <group>
+                    <field name="aws_host" />
+                    <field name="aws_access_key_id" attrs="{'required': [('backend_type', '=', 'amazon_s3')]}"/>
+                    <field name="aws_secret_access_key" attrs="{'required': [('backend_type', '=', 'amazon_s3')], 'readonly':[('backend_type', '!=', 'amazon_s3')]}"/>
+                    <field name="aws_bucket"  attrs="{'required': [('backend_type', '=', 'amazon_s3')]}"/>
+                    <field name="aws_cache_control"/>
+                </group>
+                <group>
+                        <field name="aws_region" widget="radio"
```
lmignon, line 17: @mileo IMO we should use the default selection witdget for  `aws_region` (not blocking at all)
image

lmignon (approved):
Thank your for the migration @mileo !

storage_backend_s3/components/s3_adapter.py
```diff
@@ -24,15 +24,20 @@ class S3StorageAdapter(Component):
...
         }
         if self.collection.aws_host:
             params["endpoint_url"] = self.collection.aws_host
-            # region must be excluded, otherwise endpoint is ignored
-            params.pop("region_name", None)
+
+        if self.collection.aws_region == 'other' and \
+                self.collection.aws_other_region:
+            params['region_name'] = self.collection.aws_other_region
+        elif self.collection.aws_region:
```
simahawk: if we don't have an empty option - as it seems to be - is going to break non AWS services.

simahawk (changes requested):

storage_backend_s3/components/s3_adapter.py
simahawk: AFAIU the region is needed if you rely on AWS routing, hence IMO you don't need a region if you specify a custom endpoint -> that's why I popped the region. 
Have you experienced something else?

storage_backend_s3/views/backend_storage_view.xml
mileo: Hi,

I add a new option, named "Other", that option opens a field below, that you can put in blank or specify you configuration.

For example:

```
params = {
    "aws_access_key_id": key,
    "aws_secret_access_key": secret,
    "region_name": 'fr-par',
    "endpoint_url": 'http://s3.fr-par.scw.cloud',
}

resource = boto3.resource(
    's3',
    **params,
)

for bucket in resource.buckets.all():
    print(bucket.name)
```
image

storage_backend_s3/components/s3_adapter.py
mileo: Im my example above, you can see that you can choose other region and keep the region name empty. 

Because not all non-AWS services have no defined region.

Your code don't work with the code example above.

simahawk: `aws_region` cannot be empty, look at your radio widget ;)
You are forced to select something. You should add `[('', 'None')]` to the options.

simahawk: also, I added a test for this: can you update it or add another one?

storage_backend_s3/views/backend_storage_view.xml
simahawk: yes please, rollback to selection

storage_backend_s3/components/s3_adapter.py
mileo: You are right sorry, i fixed the logic and push again. 

Please take a look.

storage_backend_s3/tests/test_fake_s3.py
```diff
@@ -0,0 +1,44 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+import logging
+import os
+from odoo.addons.storage_backend.tests.common import Common
+
+_logger = logging.getLogger(__name__)
+
+
+class FakeS3Case(Common):
```
simahawk, line 12: any reason not to update the existing one?

https://github.com/OCA/storage/blob/5c423809e64007007f5a21d06bc64974f96923f7/storage_backend_s3/tests/test_amazon_s3.py#L46

simahawk (approved):

storage_backend_s3/tests/test_fake_s3.py
lmignon, line 12: @mileo These new tests should got into the existing test class.... That'll ease the maintenance and avoid duplicate code...

simahawk: @mileo if you can take care of this we can merge :)

simahawk:
updated w/ v12 and finished here https://github.com/OCA/storage/pull/36

PR #29, opened 2019-11-18
Revert "Remove read access on storage backend for common user"

rousseldenis:
This reverts commit 22625eaaf8386ef24bec9f9e1503f467960ae0ee.

### DIFF
--- a/storage_backend/security/ir.model.access.csv
+++ b/storage_backend/security/ir.model.access.csv
@@ -1,2 +1,3 @@
 id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
 access_storage_backend_edit,storage_backend edit,model_storage_backend,base.group_system,1,1,1,1
+access_storage_backend_read,storage_backend read,model_storage_backend,base.group_user,1,0,0,0


rousseldenis:
@Cedric-Pigeon @lmignon @florian-dacosta

florian-dacosta:
Hi @rousseldenis 
Sorry, I missed your comment from 20 days ago.

It came from a discussion with @sebastienbeau, but I am not sure I remember well why we did this.
I think it is because storage is too generic and so we don't want to give rights to such a basic group.

I mean, storage backend is a generic tool to send/receive file between odoo and an external system.
You could have multiple modules using storage backend.
In this repo we have modules to store product images for example.
But we also could have a module that synchronize some bank statement files or other things more or less critical. And then, we want that a specific group is able to go and retrieve/store product image, but we do not want that same user to store/retrieve bank statement !
So, we must have a right management in these submodules, and we don't need/want manage this here.

So, the the idea, if I am correct is that these rights issues must be attended in the submodules, where the accountant manager can click on his button to retrieve the bank statement, and the rights may be bypassed in this case for this action (with sudo or suspend security).

Then we avoid that the product manager may find a way to retrieve the bank accounts files...

Would be nice to have @sebastienbeau opinion, as I am not sure this is the whole reason

rousseldenis:
IMHO, that should be the contrary. Let a read for all and when needed for security reasons on a specific module, add a record rule on backend depending on 'backend_type' field.

@lmignon

sebastienbeau:
@lmignon @rousseldenis 
Hi
We remove the default read access because this mean that every default user we be able to read the login/password of your storage backend.
I think that only "admin/responsable" user should be able to read the login/password. And normal user should be able to use the feature 'sending/getting file' by having sudo on allowed method.

What do you think? Which case is blocking for you ?

rousseldenis:
> @lmignon @rousseldenis
> Hi
> We remove the default read access because this mean that every default user we be able to read the login/password of your storage backend.
> I think that only "admin/responsable" user should be able to read the login/password. And normal user should be able to use the feature 'sending/getting file' by having sudo on allowed method.
> 
> What do you think? Which case is blocking for you ?

Using sudo is generally to be avoided.

If it is about login/password visibility, add a 'groups' attribute with 'base.group_system' field in forms for specific modules.

sebastienbeau:
It's not about visibility it's really about access of the password.
You do not want to give to any one the password of your sftp server where you have store your image, bank statement or whatever data (this information can be stolen with xmlrpc call).
But indeed blocking the full read is maybe too much.
Maybe just adding groups in field on python side is the best option?
password = fields.Char(groups='base.group_system'....)

What do you think? If we add the read access in this PR we should add this field restriction to avoid having a security issue

Cedric-Pigeon:
@rousseldenis could you include @sebastienbeau suggestion?

rousseldenis:
Ok, closing this as all calls that implements backend should be done with sudo().

PR #31, opened 2019-11-22
[12.0] Improvement and fix

simahawk:
1. S3: add option to set ACL on files
2. compute_url was not considering the directory path

### DIFF
--- a/storage_file/models/storage_file.py
+++ b/storage_file/models/storage_file.py
@@ -135,20 +135,33 @@ class StorageFile(models.Model):
             else:
                 rec.data = None
 
-    @api.depends("backend_id.served_by", "backend_id.base_url", "relative_path")
+    @api.depends(
+        "backend_id.served_by",
+        "backend_id.base_url",
+        "backend_id.url_include_directory_path",
+        "relative_path",
+    )
     def _compute_url(self):
         for record in self:
-            if record.backend_id.served_by == "odoo":
-                base_url = (
-                    self.env["ir.config_parameter"].sudo().get_param("web.base.url")
-                )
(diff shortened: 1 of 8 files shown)


guewen (approved):

storage_file/models/storage_file.py
```diff
@@ -146,9 +146,11 @@ def _compute_url(self):
                     base_url, record._slugify_name_with_id()
                 )
             else:
-                record.url = "{}/{}".format(
-                    record.backend_id.base_url, record.relative_path
-                )
+                parts = [record.backend_id.base_url]
```
lmignon: @simahawk You are here on a point discussed a lot of times with @Cedric-Pigeon. His POV is that the `directory_path` is the internal path used to store you image into the backend. The url should not use this information. If the file is not served at the root of `base_url`, it's your responsibility to put  in place a redirection or configure a CDN (for S3)

If you use a CDN into S3 you can use the `shopivader_backend_image_proxy` to rewrite this URL...  when the info is exported to alglio.... 

I'm not convince by this sapproach and we can discuss about it with @Cedric-Pigeon and @sebastienbeau

simahawk: I'll check shopivader_backend_image_proxy.
hmm then we could make it optional because w/ exoscale by default it won't work.
What if I add a param to include it or not then every one is free to use it or not, fine?
I mean, this is not only for images and should not depend on shopinvader features of course.

simahawk:
@lmignon what about now? I added tests for it ;)

simahawk:
fixed and fixed a runbot warning as well.

PR #32, opened 2019-11-22
Add support for server_env conf

simahawk:


storage_file/models/storage_file.py
```diff
@@ -190,3 +190,13 @@ def get_from_slug_name_with_id(self, slug_name_with_id):
...
         return self.browse(_id)
+
+    def _get_backend_id_from_param(self, param_name):
+        param = self.env["ir.config_parameter"].sudo().get_param(param_name)
+        if param:
+            if param.isdigit():
+                return int(param)
+            elif "." in param:
+                # assume is a xid, let it raise if not found
+                return self.env.ref(param)
```
lmignon:
```suggestion
                return self.env.ref(param).id
```

lmignon (commented):
@simahawk Great idea!!! but travis is red :-(

simahawk:
@lmignon fixed and:
* added specific test for param retrieval
* unified file type handling
* unified and renamed backend retrieval

simahawk:
ouch, I've left `aws_file_acl` into the computed fields :/
I'll get back to this PR as https://github.com/OCA/storage/pull/31 is merged

simahawk:
includes #31 now to not forget about new fields

simahawk:
have to solve some weird inheritance issues. I'll look into it

storage_file/models/storage_backend.py
```diff
@@ -37,3 +39,33 @@ class StorageBackend(models.Model):
         "Public: your file/image can be displayed if nobody is "
         "logged (useful to display files on external websites)",
     )
+    url_include_directory_path = fields.Boolean(
```
lmignon, line 42: :+1:

storage_file/models/storage_file.py
```diff
@@ -135,20 +135,33 @@ def _compute_data(self):
...
                 rec.data = None
 
-    @api.depends("backend_id.served_by", "backend_id.base_url", "relative_path")
+    @api.depends(
+        "backend_id.served_by",
+        "backend_id.base_url",
+        "backend_id.url_include_directory_path",
+        "relative_path",
+    )
     def _compute_url(self):
```
lmignon, line 138: @simahawk Are we sure that compute methods are triggered if the values provided by the `server.env.mixin` changes?

lmignon (approved):
:heart_eyes: Thank you @simahawk (Code review)

storage_file/models/storage_file.py
simahawk: hmm very good point. @guewen what is the recommended way in this case?
Context: this is a computed + stored field depending on server env params.

lmignon: @simahawk IMO it's not a blocking point. It would probably be necessary to add a warning in the README on this point and propose a procedure to launch this recomputation manually (via click-odoo or ????) if these values are modified.

simahawk: not sure. Can be risky whenever you move a db from prod to integration for instance.
Would be nice to have a common well-known solution.

guewen: It kind of defeats the purpose of environment, since a copy of a prod database to another environment will not reset the url.

Does it need to be stored?

simahawk: I don't think so. This computation is not so heavy and is not used so often AFAIR.
OTOH can be done on potentially gazillions of records (images + their scales).

lmignon: That's the point. I share the same POV. The url should no more be stored...  If I remember well, into the past, the url field was used by a controller to lookup for the record... it's no more the case and it seems safe to no more store this value into the database.

simahawk:
argh! breaking tests now :sob: Checking...

lmignon:
:champagne: :heart: :tada:

PR #33, opened 2019-11-26
storage_image_product: add wizard to migrate product images

simahawk:
Comes from a post init hook by @lmignon 
Not really tested yet.

### DIFF
--- a/storage_image_product/__init__.py
+++ b/storage_image_product/__init__.py
@@ -1,1 +1,2 @@
 from . import models
+from . import wizards
--- a/storage_image_product/__manifest__.py
+++ b/storage_image_product/__manifest__.py
@@ -20,5 +20,6 @@
         "views/product_product.xml",
         "views/product_category.xml",
         "views/image_tag.xml",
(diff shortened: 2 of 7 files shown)


lmignon (commented):
Thank you @simahawk I'll try to test it in the coming days.

rousseldenis:
@simahawk What's the status of this ?

simahawk:
@rousseldenis currently not used on our side. It's a nice to have (of must have for existing instances w/ images) and if someone could test and write tests it would be nice. I'd keep it open or reference it in an issue.

lmignon:
@rousseldenis This one provide a wizard that can be used to migrate existing product images when the addon is installed on an existing DB. It should be finalized.... but as always that require some times to do it.

codeagencybe:
Is this feature working and available? 
And does it only migrate images or also other attachments from records?

lmignon:
@codeagencybe This feature is only available into the branch linked to the PR. The wizard performs the migration of the product.template's images only.

PR #34, opened 2020-01-10
storage_image: improve tests common

simahawk:
Overriding the base test user is bad if you want to inherit from the common case and test different things. Admin user has been assigned explicitly by the module to the right group.
If you want to test rights, then add a proper test for it.


simahawk:
@lmignon do you mind if I fast track this?

PR #35, opened 2020-01-14
[12.0][IMP] storage_backend, filter files in _list by pattern, add notebook to storage backend view

GSLabIt:
@florian-dacosta can u double check?

### DIFF
--- a/storage_backend/models/storage_backend.py
+++ b/storage_backend/models/storage_backend.py
@@ -2,9 +2,10 @@
 # @author Sébastien BEAU <sebastien.beau@akretion.com>
 # Copyright 2019 Camptocamp SA (http://www.camptocamp.com).
 # @author Simone Orsi <simone.orsi@camptocamp.com>
+# Copyright 2020 Giovanni Serra
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
-
 import base64
+import fnmatch
 import logging
 
 from odoo import fields, models
@@ -42,8 +43,11 @@ class StorageBackend(models.Model):
     def _get_bin_data(self, relative_path, **kwargs):
         return self._forward("get", relative_path, **kwargs)
 
-    def _list(self, relative_path=""):
-        return self._forward("list", relative_path)
+    def _list(self, relative_path="", pattern=False):
+        names = self._forward("list", relative_path)
+        if pattern:
+            names = fnmatch.filter(names, pattern)
+        return names
 
     def _delete(self, relative_path):
         return self._forward("delete", relative_path)
--- a/storage_backend/README.rst
+++ b/storage_backend/README.rst
@@ -57,6 +57,7 @@ Contributors
 * Benoît Guillot <benoit.guillot@akretion.com>
 * Laurent Mignon <laurent.mignon@acsone.eu>
 * Denis Roussel <denis.roussel@acsone.eu>
+* Giovanni Serra <giovanni@gslab.it>
 
 Maintainers
 ~~~~~~~~~~~
--- a/storage_backend/components/filesystem_adapter.py
+++ b/storage_backend/components/filesystem_adapter.py
@@ -1,7 +1,6 @@
 # Copyright 2017 Akretion (http://www.akretion.com).
 # @author Sébastien BEAU <sebastien.beau@akretion.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
-
 import os
 
 from odoo import _
--- a/storage_backend/views/backend_storage_view.xml
+++ b/storage_backend/views/backend_storage_view.xml
@@ -26,6 +26,7 @@
                         <field name="backend_type"/>
                         <field name="directory_path"/>
                     </group>
+                    <notebook name="backend_config" attrs="{'invisible': [('backend_type', '=', False)]}" />
                 </sheet>
             </form>
         </field>


GSLabIt:
@pedrobaeza @bguillot sorry for bothering, why travis is failing?

florian-dacosta:
@GSLabIt 
I think the build fails because you did not use pre-commit, and it is required for this repo.
I am not sure if there is documentation somewhere.
You should install pre-commit, with pip for instance, and then in the repo, run `pre-commit install`

Then you can run `pre-commit run --all`
What  seems to have fails is isort, it is used to sort the import.
the `import fnmatch` should probably go before the `import logging`

I leave someone who know more about pre-commit correct me if I am wrong.

florian-dacosta (approved):
Seems good to me.
@sebastienbeau @lmignon @Cedric-Pigeon Could you make a quick review please?

GSLabIt:
could this be merged?

florian-dacosta:
Hi @GSLabIt 
I am not sure why you modify the storage backend view here, but I think it is not for the same purpose of your first commit.
To accelerate/maximise the merge of the PR, I think you should do one PR per feature.
So one PR can be approved and merged even if an is refused/not reviewed.
My  2 cents

GSLabIt:
@florian-dacosta i've changed to obtain this:

Schermata 2020-01-20 alle 18 25 39

It will enable the ability to inherit notebook from `storage_backand_*` implementation. Thought it would be a nice and clean look and feel instead of the actual and no big view changes just 1 line ..Don't think time to merge will be influenced by this ...

storage_backend/models/storage_backend.py
```diff
@@ -42,8 +43,11 @@ def _add_bin_data(self, relative_path, data, **kwargs):
     def _get_bin_data(self, relative_path, **kwargs):
         return self._forward("get", relative_path, **kwargs)
 
-    def _list(self, relative_path=""):
-        return self._forward("list", relative_path)
+    def _list(self, relative_path="", pattern=False):
```
lmignon, line 46: @GSLabIt Thank you for your proposal.
I don't see any benefit to put this filter pattern into the core api. That being said, if we want to provide this functionality into the API, this information should be propagated to the specific backend adapter implementation.  I've the feeling that in some cases the filtering could be done by the specific backend...

lmignon (changes requested):

GSLabIt:
Well, I followed hint form @florian-dacosta to put that at this level. 

Btw, what's best solution to have file filtering? Fw to specific backend or other?

storage_backend/models/storage_backend.py
GSLabIt: Well, I followed hint form @florian-dacosta to put that at this level.

Btw, what's best solution to have file filtering? Fw to specific backend or other?

lmignon: IMO forward tot specific backend

florian-dacosta: Thanks for the feedback @lmignon 
I've discussed it a little with @sebastienbeau 

The thing is, at first sight, it does not seem that specific backend lib like boto3 or paramiko provide this kind of possibility.
That is why we wanted to put this in code api and not in specific backend.

It we have to implement it in all our specific backend, we will just be duplicating code no?
Or what would you suggest?

Also I wonder, why wouldn't you think it is a good feature to be at this level?
I believe that if we want the storage backend modules to be wildely used, it should be generic enough to manage multiple use cases, and wanting to read only specific files from a external storage seems very generic!

GSLabIt: oh ok ..i better wait for a final decision then .. 😬

lmignon: @florian-dacosta I understand your point but it's not because some specific backends doesn't provide this functionality that we must prevent the implementation of this feature at the level of a specific backend. It's true that the implementation at the specific backend layer could introduce some duplicate code. But this drawback can be mitigated by implementing an helper method used by the specific backends that are not able to natively implement this functionality.

GSLabIt: @lmignon at the end of the day it will be a filter applied to a list of file names, so it not depends on backend; that's why implementing at this level.

florian-dacosta: Yes, what I thought was to do it this way until we have a specific backend providing this feature, and this day, if he arrives, doing a refactore. Indeed, I think we often do complicated stuff in case one day...that may never come!

One important thing is that we want to be sure that this feature is implemented for all backends.
Our use case with the module attachment_synchronize is to get the file names respecting a given pattern. But, this module may use any specific backend implemented.
So, if `storage_backend_sftp` implement it, but not `storage_backend_ftp` for example, it will fail in the second case.

I don't really see a very good solution.

So, what could be done : 
Change the list method this way : 
```
    def _list(self, relative_path="", pattern=""):
        file_names =  self._forward("list", relative_path, pattern=pattern)
        file_names = self._forward("filter_filenames", relative_path, file_names, pattern)
```
Then, define `filter_filenames` method to do the filtering in `BaseStorageAdapter` componenent.
Then, if a specific backend want to implement it another way, it can do it by overriding `filter_filenames` so it does not do anything...

Second solution which is not good enough IMO, because one could forget to implement it for a specific backend, which would result on possible bugs for submodules assuming any backend allow this.

Change le list method this way : 
```
    def _list(self, relative_path="", pattern=""):
        return self._forward("list", relative_path, pattern=pattern)
```
in `BaseStorageAdapter` add an helper method that take a list of name and filter it.
Then, in each specific backend component, use the helper in the list method.
```
def list(self, relative_path="", pattern=""):
    filenames = ...
    filenames = self.helper_method(filenames, pattern)
```

If a new specific backend is implemented, we can choose to use the helper or implement this another way.

I am not sure any of this is the best option.
What do you think @lmignon

GSLabIt: > Change the list method this way :
> 
> ```
>     def _list(self, relative_path="", pattern=""):
>         file_names =  self._forward("list", relative_path, pattern=pattern)
>         file_names = self._forward("filter_filenames", relative_path, file_names, pattern)
> ```

Not getting this u mean change `_forward list` to `_forward filter_filenames`? 

I still think filtering a list would not depend on storage backend. As far as i know, in ftp, sftp, it's a common solution to filter list on `listdir` results, because it take only `path` as parameter.

What about adding a new api method in `storage_backend` kind of:
 ```
     def _filtered_list(self, relative_path="", pattern=""):
         file_names =  self._forward("list", relative_path)
         if pattern:
              file_names = fnmatch.filter(file_names, pattern)
```

lmignon: > Yes, what I thought was to do it this way until we have a specific backend providing this feature, and this day, if he arrives, doing a refactore. Indeed, I think we often do complicated stuff in case one day...that may never come!

OK for me.

lmignon (approved):

GSLabIt:
merge?

PR #36, opened 2020-01-16
[13.0][MIG] storage_backend_s3

simahawk:
Replaces #26 
Includes a merge from 12.0


simahawk:
@lmignon good for you?

lmignon (approved):
LGTM! Thank you @simahawk

storage_backend_sftp/__manifest__.py
```diff
@@ -5,7 +5,7 @@
 {
     "name": "Storage Backend SFTP",
     "summary": "Implement SFTP Storage",
-    "version": "12.0.1.0.0",
```
mileo, line 8: 13?

mileo (approved):

storage_backend_sftp/__manifest__.py
simahawk: this module is not migrated yet, it comes from 12 as it is

PR #45, opened 2020-03-24
[12.0][IMP] storage_backend_s3 - add compatibility with OVH's Openstack

hparfr:
Add support of regions other than ones from AWS (manually entered by the user).

In the case of Openstack / OVH , region and enpoint (host) should be used.
@sebastienbeau do you know in which case region should be removed ?

### DIFF
--- a/storage_backend_s3/models/storage_backend.py
+++ b/storage_backend_s3/models/storage_backend.py
@@ -6,7 +6,7 @@
 
 import logging
 
-from odoo import fields, models
+from odoo import fields, models, api
 
 _logger = logging.getLogger(__name__)
 
@@ -23,14 +23,13 @@ class StorageBackend(models.Model):
     aws_host = fields.Char(
         string="AWS Host",
         help="If you are using a different host than standard AWS ones, "
-        "eg: Exoscale",
+        "eg: Exoscale, Openstack. Also called endpoint",
     )
     aws_bucket = fields.Char(string="Bucket")
     aws_access_key_id = fields.Char(string="Access Key ID")
     aws_secret_access_key = fields.Char(string="Secret Access Key")
-    aws_region = fields.Selection(
-        selection='_selection_aws_region', string="Region"
-    )
+    aws_region = fields.Char(string="Region",
+        help="Can be empty for some providers")
     aws_cache_control = fields.Char(default="max-age=31536000, public")
     aws_file_acl = fields.Selection(selection=[
         ('', ''),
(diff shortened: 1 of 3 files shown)


sebastienbeau:
@hparfr I think it's in case of Exoscale @simahawk have done the change here https://github.com/OCA/storage/commit/cdc4f2d5c7016c025c571db703c9410fb35a9ec7

sebastienbeau (approved):
LGTM (code review).
Thanks for helping to support more S3 alternative

storage_backend_s3/models/storage_backend.py
```diff
@@ -58,8 +61,17 @@ def _server_env_fields(self):
...
+        # it introduce latency in rendering of
+        # form view and tree view of storage.backend
         session = boto3.session.Session()
         return [
             (region, region.replace("-", " ").capitalize())
             for region in session.get_available_regions("s3")
         ]
+
+    @api.onchange("aws_available_regions")
+    def _set_region(self):
```
simahawk: by convention this should be `_onchange_available_regions`

storage_backend_s3/models/storage_backend.py
```diff
@@ -58,8 +61,17 @@ def _server_env_fields(self):
...
 
     def _selection_aws_region(self):
+        # This calls AWS web servers,
+        # it introduce latency in rendering of
+        # form view and tree view of storage.backend
         session = boto3.session.Session()
         return [
             (region, region.replace("-", " ").capitalize())
             for region in session.get_available_regions("s3")
         ]
```
simahawk: I think it's better to store this in model cache instead.

```suggestion
    _aws_available_regions = None
    
    def _selection_aws_region(self):
        if self._aws_available_regions:
            return self._aws_available_regions
        # This calls AWS web servers,
        # it introduce latency in rendering of
        # form view and tree view of storage.backend
        session = boto3.session.Session()
        self._aws_available_regions = [
            (region, region.replace("-", " ").capitalize())
            for region in session.get_available_regions("s3")
        ]
        return self._aws_available_regions
```

simahawk (changes requested):
wondering: shall we add a `aws_provider` to have unique ways of handling params punctually?

PierrickBrun (approved):

simahawk (approved):
@hparfr can you squash?

simahawk:
> wondering: shall we add a `aws_provider` to have unique ways of handling params punctually?

BTW any feedback on this?

hparfr:
> wondering: shall we add a aws_provider to have unique ways of handling params punctually?

I don't think we need it now. We will see if other s3-like implementation specifics arise in the future. 


>@hparfr can you squash?
Can't we use squash on merge ?

simahawk:
not if we use ocabot as we usually do ;)

PR #48, opened 2020-05-27
[12.0][ADD] Storage image product: Image handle widget

qgroulard:
Adding new images on products is tedious since you need to manually create a `product_image_relation` record and  a `storage_image` linked to it for each image you want to add on a product.
Furthermore, there is currently no way to change images order (except by removing an image and re-adding it).

This widget allows to directly drop image files in the 'Image' tab on the product form. Related `product_image_relation` and  `storage_image` will automatically be created.
You can drop multiple images at once.
Plus, images cards in the 'Image' tab can now be reorganized by drag-and-drop.


apineux (approved):

sebastienbeau:
Great work.

Some improvement that we can do (I can do it if you want)

We can add a mixin for the *.image.relation
the following code is really similare: 
- https://github.com/OCA/storage/blob/12.0/storage_image_product/models/category_image_relation.py#L13
- https://github.com/OCA/storage/blob/12.0/storage_image_product/models/product_image_relation.py#L13

so we can have a "image.relation.mixin" in the module storage_image and we can move the code of your widget into storage_image.

By doing this it will be easier to reuse your widget is other case.
Also in the module storage_image_product we should add you widget for the category here : 
- https://github.com/OCA/storage/blob/12.0/storage_image_product/views/product_category.xml#L10


After we have merge this PR, we can take some time to improve the UI of the pop-up when we add a image manually.

storage_image_product/static/src/js/storage_image_product.js
```diff
@@ -0,0 +1,100 @@
+odoo.define('storage_image_product.image_handle', function (require) {
```
rousseldenis: @qgroulard Can't this widget be added on storage_image module level instead of product one ? In order to be reused.

rousseldenis: Arf, didn't read @sebastienbeau comment.

qgroulard:
> Great work.
> 
> Some improvement that we can do (I can do it if you want)
> 
> We can add a mixin for the *.image.relation
> the following code is really similare:
> 
>     * https://github.com/OCA/storage/blob/12.0/storage_image_product/models/category_image_relation.py#L13
> 
>     * https://github.com/OCA/storage/blob/12.0/storage_image_product/models/product_image_relation.py#L13
> 
> 
> so we can have a "image.relation.mixin" in the module storage_image and we can move the code of your widget into storage_image.
> 
> By doing this it will be easier to reuse your widget is other case.
> Also in the module storage_image_product we should add you widget for the category here :
> 
>     * https://github.com/OCA/storage/blob/12.0/storage_image_product/views/product_category.xml#L10
> 
> 
> After we have merge this PR, we can take some time to improve the UI of the pop-up when we add a image manually.

Thanks for the feedback.
I've started the mixin (an abstract class over *.image.relation classes for now), but I'm not sure of how the rest is specific (available attributes and tags). What are those fields used for ?

Feel free to finish the job if you want to, or I'll do it once I understand better the relation objects.

I've moved everything to storage_image and added the widget on the categories.

rousseldenis (approved):
LGTM

lmignon (approved):

lmignon:
@sebastienbeau Can you update your review PLZ

sebastienbeau (approved):
LGTM

PR #44, opened 2020-03-23
[12.0] storage_file: fix error creating a new file

hparfr:

PR #49, opened 2020-05-29
[13.0] storage_file: fix error creating a new file

amon-ra:
This patch fixes an error when a new file is created using the backend ui.

It is very simple and self explanatory.

It is the same as: GH-44 but for 13.0


simahawk (changes requested):
#44 is getting merged. We should forward port that changes because we want test and keep changelog consistent.

sebastienbeau:
I close it as the forward port is here : https://github.com/OCA/storage/pull/71

PR #55, opened 2020-06-17
[13.0][ADD] Image handle widget

qgroulard:
Forward port of https://github.com/OCA/storage/pull/48

Without the drag&drop reordering part since it's now supported by Odoo. (https://github.com/odoo/odoo/pull/28581/commits/2c28ffe258fa470e79f3eb786f0cb76eb8f20dcd)
We combine this widget with the widget 'handle' on the sequence field to achieve the same result as in https://github.com/OCA/storage/pull/48.

### DIFF
--- a/storage_image/models/__init__.py
+++ b/storage_image/models/__init__.py
@@ -1,2 +1,3 @@
 from . import storage_image
 from . import storage_file
+from . import storage_relation_abstract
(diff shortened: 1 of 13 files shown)


sebastienbeau (approved):
Already approuved on V12, LGTM

PR #57, opened 2020-06-19
[13.0] Add generic validation + sftp fix/imp

simahawk:
Commits are self explanatory.

### DIFF
--- a/storage_backend/models/storage_backend.py
+++ b/storage_backend/models/storage_backend.py
@@ -7,7 +7,7 @@
 import base64
 import logging
 
-from odoo import fields, models
+from odoo import _, fields, models
 
 _logger = logging.getLogger(__name__)
 
@@ -25,6 +25,12 @@ class StorageBackend(models.Model):
     directory_path = fields.Char(
         help="Relative path to the directory to store the file"
     )
+    has_validation = fields.Boolean(compute="_compute_has_validation")
+
+    def _compute_has_validation(self):
+        for rec in self:
+            adapter = self._get_adapter()
+            rec.has_validation = hasattr(adapter, "validate_config")
 
     @property
     def _server_env_fields(self):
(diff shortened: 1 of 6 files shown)


storage_backend/models/storage_backend.py
```diff
@@ -5,13 +5,38 @@
...
+import inspect
 import logging
 
-from odoo import fields, models
+from odoo import _, fields, models
 
 _logger = logging.getLogger(__name__)
 
 
+def _check_class_direct_func(instance, func_name):
```
simahawk: @guewen do you know how to work around this?
I need to check if the method is re-defined at component level but both approaches do not work.
I could call the method and check for NotImplementedError but I don't want to call it as depending on the service it might hang during computation.

guewen: Couldn't the components implement a method with a different name? If it exists, use it, otherwise, use the generic one?

simahawk: yes but then I can also not define the method on the base adapter, that would be easier.
It's just that I'm trying to keep it consistent w/ the rest and understand what's going on.
To me, the class attached to the component is wrong or at least the module is wrong, no?

guewen: All components are registered in `odoo.addons.component.core` the same way all models are registered in `odoo.api`:

```
(Pdb) pp self.env["res.partner"].__class__
<class 'odoo.api.res.partner'>
(Pdb) pp self.env["sale.order"].__class__
<class 'odoo.api.sale.order'>
```

The thing is: these classes are like virtual classes which are built by aggregating all the classes found for a model/component. So they don't really have a proper module. We could have used the module of the first class found with `_name`, but I don't think we should handle differently than odoo, and it would maybe be even more confusing, because, for instance given the class FoorBar exist in module `odoo.addons.connector_foo`, the actual composite class `FooBar` built from all the classes found would not really be equal to the class `odoo.addons.connector_foo.FooBar`.

simahawk: ok, tnx. I've opted for the easy way: not define the method on the base class.

guewen (approved):

simahawk:
@acsonefho could you review this? :pray:

acsonefho (approved):
LGTM (code only, not tested)

PR #65, opened 2020-09-09
[12.0][FIX] storage_image: Indefinite error

rousseldenis:
When the url is not accessible, the error event on
FieldImageUrl object is launched indefintely

Before:

image

After:

image


rousseldenis:
@sebastienbeau

lmignon (approved):

hparfr (approved):

sebastienbeau (approved):
+1000 I hate this error !

PR #74, opened 2020-11-23
[12.0][IMP] improve perf by computing all field in the same time

sebastienbeau:


simahawk (approved):

sebastienbeau:
In fact I do not have the right invalidation
if the url change (backend storage config change) it will not propagate it correctly

JordiBForgeFlow:
@sebastienbeau I did something very similar here: https://github.com/OCA/storage/pull/84
What I added on top of your idea is to consider the sorting of the images in the recomputation. Otherwise the image that appears in the product as main image is not necessarily the first in the sequence.

MiquelRForgeFlow:
could you consider https://github.com/OCA/storage/pull/84 instead, as it include more things?

Issue #51, opened 2020-06-04
storage backend public api


simahawk:
`storage.backend` api is not public and to me makes little sense since to use the storage you must call those methods like `_add_b64_data` `_ad_bin_data` etc by guessing content encoding.

We should an api similar to adapters and call b64 or bin based on the input.
The final API would be:

* add (or put or similar)
* get (or read, because it returns the content of the file)
* list (or list_dir or dir or similar)
* delete

simahawk:
implemented BTW

PR #75, opened 2020-11-24
Implement public API

simahawk:
Implements https://github.com/OCA/storage/issues/51

### DIFF
--- a/storage_backend/models/storage_backend.py
+++ b/storage_backend/models/storage_backend.py
@@ -5,11 +5,52 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 import base64
+import functools
+import inspect
 import logging
+import warnings
 
 from odoo import _, fields, models
 
 _logger = logging.getLogger(__name__)
+
+
+# TODO: useful for the whole OCA?
+def deprecated(reason):
+    """Mark functions or classes as deprecated.
+
+    Emit warning at execution.
+
+    The @deprecated is used with a 'reason'.
+
+        .. code-block:: python
+
+            @deprecated("please, use another function")
+            def old_function(x, y):
+                pass
+    """
+
+    def decorator(func1):
+
+        if inspect.isclass(func1):
+            fmt1 = "Call to deprecated class {name} ({reason})."
+        else:
+            fmt1 = "Call to deprecated function {name} ({reason})."
+
+        @functools.wraps(func1)
+        def new_func1(*args, **kwargs):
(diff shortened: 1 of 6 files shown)


storage_backend/models/storage_backend.py
```diff
@@ -5,13 +5,54 @@
...
+import inspect
 import logging
+import warnings
 
 from odoo import _, fields, models
 
 _logger = logging.getLogger(__name__)
 
 
+# TODO: useful for the whole OCA?
```
sebastienbeau, line 18: yes this should be move in a generic tool !

sebastienbeau:
@simahawk Initialy I choose to not have any public api on storage backend in order to avoid exposing this method to the xmlrpc webservice and so to reduce potential security issue (maybe it's useless)
It's will be interesting to have @lmignon opinion

lmignon (approved):
I like this clean API. Regarding security, indeed it's painful that any non protected methods are available through XML/JSONRPC calls. IMO it's against the python language best practices but we can't change this. In every cases, the security is enforced by ACLs and Record Rules.

simahawk:
> I like this clean API. Regarding security, indeed it's painful that any non protected methods are available through XML/JSONRPC calls. IMO it's against the python language best practices but we can't change this. In every cases, the security is enforced by ACLs and Record Rules.

Same feeling here..

FWIW I've tried to propose this https://github.com/odoo/odoo/issues/62299

sebastienbeau (approved):
Thanks for your feedback.
It's seem cleaner to later add @api.rpc_disable (if exist)
In any case if the storage backend have the right access right (no read access for unwanted user) the API will be not available.

sebastienbeau:
@simahawk do you have any plan of other refactor in V13 for storage (or search engine) ? I plan to migrate this repo this end of week

storage_backend/models/storage_backend.py
simahawk: suggestions?
Shall we embrace the proposal by Pedro to have an all in one `odoo-oca-lib`?

simahawk:
> @simahawk do you have any plan of other refactor in V13 for storage (or search engine) ? I plan to migrate this repo this end of week

not that I'm aware of :rofl: :stuck_out_tongue:

storage_backend/models/storage_backend.py
lmignon: @simahawk Personally I'm against a all in one lib. I prefer a small (even very small) lib with a dedicated scope than one with tons of unrelated stuffs.

PR #81, opened 2021-01-26
[13.0][storage_image_product][fix] sort order to calculate the default image

JordiBForgeFlow:
Fixes the sort order of images in the product template and variant, used
to compute the default image displayed in the product.

Before this fix, having an image in the first position in the "Images" tab of the product would not guarantee that this image would be used as the default image of the product.

image


lmignon:
@JordiBForgeFlow Can you force push your branch to trigger a new build based on 13.0?

PR #83, opened 2021-01-26
[13.0][storage_image_product] add tech_name field

JordiBForgeFlow:
Add the field 'tech_name' to better manage selection of product images by tag, instead of using the normal name.

### DIFF
--- a/storage_image_product/migrations/13.0.1.3.0/post-migration.py
+++ b/storage_image_product/migrations/13.0.1.3.0/post-migration.py
@@ -1,0 +1,10 @@
+# Copyright 2021 ForgeFlow S.L. (https://www.forgeflow.com)
+# Part of ForgeFlow. See LICENSE file for full copyright and licensing details.
+from odoo import SUPERUSER_ID, api
+
+
+def migrate(cr, version):
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    tags = env["image.tag"].search([])
+    for tag in tags:
+        tag.tech_name = env["image.tag"]._normalize_tech_name(tag.name)
--- a/storage_image_product/models/image_tag.py
+++ b/storage_image_product/models/image_tag.py
@@ -8,6 +8,7 @@ from odoo import api, fields, models
 
 class ImageTag(models.Model):
     _name = "image.tag"
+    _inherit = ["server.env.techname.mixin"]
     _description = "Image Tag"
 
     @api.model
--- a/storage_image_product/views/image_tag.xml
+++ b/storage_image_product/views/image_tag.xml
@@ -5,6 +5,7 @@
         <field name="arch" type="xml">
             <tree string="Image Tag" editable="bottom">
                 <field name="name" />
+                <field name="tech_name" />
                 <field name="apply_on" />
             </tree>
         </field>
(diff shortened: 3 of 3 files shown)


storage_image_product/__manifest__.py
```diff
@@ -5,7 +5,7 @@
 {
     "name": "Storage Image Product",
     "summary": "Link images to products and categories",
-    "version": "13.0.1.2.0",
+    "version": "13.0.1.3.0",
```
MiquelRForgeFlow: ocabot automatically does this

MiquelRForgeFlow:
This PR is targeting branch v14. Are you sure?

storage_image_product/models/image_tag.py
```diff
@@ -22,7 +24,26 @@ def _get_default_apply_on(self):
...
+    @api.model
+    def create(self, vals):
+        self._handle_tech_name(vals)
+        return super().create(vals)
+
+    def write(self, vals):
+        self._handle_tech_name(vals)
+        return super().write(vals)
+
+    def _handle_tech_name(self, vals):
```
simahawk: hmm this code looks familiar :thinking: :smile: 

@JordiBForgeFlow after re-implementing it here and there, in the ctx of server env I've added [this mixin](https://github.com/OCA/server-env/blob/13.0/server_environment/models/server_env_tech_name_mixin.py) and removed duplicated code from shopinvader and search engine modules.
Maybe it's time to make it a generic mixin?

JordiBForgeFlow: But then you need to add an extra dependency, just for this, right? IMHO in this case it's not really justified. I just used that because it was nice. I mean, if this feature was provided by a pip module, then I would agree to reuse.

simahawk: storage_backend already depends on server env https://github.com/OCA/storage/blob/13.0/storage_backend/__manifest__.py#L15

The mixin can be kept there. Also, this feature is something that it should be required for product attributes as well because they don't have unique key and the filter names in search engine keys may vary from lang to lang.

simahawk: BTW you can already use PIP to install modules :wink:

storage_image_product/migrations/13.0.1.3.0/post-migration.py
```diff
@@ -0,0 +1,11 @@
+# Copyright 2021 ForgeFlow S.L. (https://www.forgeflow.com)
+# Part of ForgeFlow. See LICENSE file for full copyright and licensing details.
+
+from openupgradelib import openupgrade  # pylint: disable=W7936
```
lmignon: @JordiBForgeFlow `openupgradelib`should be declared as the external dependency into the __manifest__.py file.

Nevertheless, do we really need an other dependecy? We can achieve the same result with the same number of lines of code by using the odoo api...
 

```python

from odoo import SUPERUSER_ID, api

def migrate(cr, version)
    env = api.Environment(cr, SUPERUSER_ID, {}) 

```

JordiBForgeFlow:
@simahawk @lmignon Thanks for your inputs! I have now attended your comments.

MiquelRForgeFlow (approved):
I think it's ready to be merged.

lmignon (approved):

PR #85, opened 2021-01-27
[13.0][imp][storage_file] create the image url using report.url if the image is being printed from an odoo report

JordiBForgeFlow:
Otherwise the report will not be able to display the image, because it is using the base.web_url instead.


simahawk (commented):
@JordiBForgeFlow as I'm improving and refactoring `_get_url` in https://github.com/OCA/storage/pull/88 and this PR would clash, I blindly added the last commit to satisfy your need. 
Can you check if it looks good for you?

simahawk:
addressed in https://github.com/OCA/storage/pull/88

PR #87, opened 2021-02-03
storage_s3: fix aws regions -> load once

simahawk:
Every time the selection field was loaded
S3 API were called to retrieve regions.
This can have a big impact on computing URLs
and in any case is useless if we assume
that regions do not change every now and then.

### DIFF
--- a/storage_backend_s3/models/storage_backend.py
+++ b/storage_backend_s3/models/storage_backend.py
@@ -16,6 +16,20 @@ except ImportError as err:  # pragma: no cover
     _logger.debug(err)
 
 
+def _load_aws_regions():
+    _logger.info("Loading available AWS regions")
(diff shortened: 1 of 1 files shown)


guewen (approved):

simahawk:
linting to be fixed here https://github.com/OCA/storage/pull/90

storage_backend_s3/models/storage_backend.py
```diff
@@ -16,6 +16,20 @@
     _logger.debug(err)
 
 
+def _load_aws_regions():
+    _logger.info("Loading available AWS regions")
+    session = boto3.session.Session()
```
lmignon, line 21: @simahawk  is the session auto closed?

lmignon (approved):
LGTM (code review only)

PR #88, opened 2021-02-03
storage_file: boost url compute

simahawk:
URL is always computed because some parts of it might change.

Reading backend data that might come form server env params
or computing all the times the slug using regex can be very consuming.

Certain bits can be stored on their respective records.

### DIFF
--- a/storage_file/models/ir_actions_report.py
+++ b/storage_file/models/ir_actions_report.py
@@ -1,0 +1,14 @@
+# Copyright 2021 Camptocamp SA (http://www.camptocamp.com).
+# @author Simone Orsi <simahawk@gmail.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class IrActionsReport(models.Model):
+    _inherit = "ir.actions.report"
+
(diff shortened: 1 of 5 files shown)


simahawk:
One more thing to consider: "served_by", "base_url", "directory_path", "url_include_directory_path" come from the env.
Does it make sense to have `@api.depends` on them?
This will still trigger the call to serve env machinery and, when served from odoo, the call to get_param.

simahawk:
last commit is a blind porting and improvement of #85

storage_file/models/ir_actions_report.py
```diff
@@ -0,0 +1,14 @@
...
+
+from odoo import models
+
+
+class IrActionsReport(models.Model):
+    _inherit = "ir.actions.report"
+
+    def render_qweb_pdf(self, res_ids=None, data=None):
+        return super(
+            IrActionsReport, self.with_context(print_report_pdf=True)
```
simahawk, line 13: this key should be probably proposed to odoo core?

lmignon (approved):
LGTM (Code review)

simahawk:
@JordiBForgeFlow good for you?

i-vyshnevska (approved):

PR #90, opened 2021-02-04
[13.0] Sync with OCA/oca-addons-repo-template 

simahawk:


.copier-answers.yml
```diff
@@ -0,0 +1,13 @@
+# Do NOT update manually; changes here will be overwritten by Copier
+_commit: v1.1.3
+_src_path: https://github.com/OCA/oca-addons-repo-template.git
+dependency_installation_mode: PIP
+generate_requirements_txt: true
```
simahawk, line 5: @sbidoul this flag hasn't produced the pre-commit entry as per https://github.com/OCA/oca-addons-repo-template/blob/2a361ba50312c50ad9e8d3404c37995a9bef212d/.pre-commit-config.yaml.jinja#L130

sbidoul: Indeed. I can reproduce but I don't understand the reason. @Yajo would you have an idea here ?

MiquelRForgeFlow (approved):

MiquelRForgeFlow:
Oh, I thought you were wainting for something

.copier-answers.yml
yajo: Yes, that file is only for v14+ branches.

PR #94, opened 2021-03-31
storage_thumbnail: do not store url

simahawk:
The URL for all storage files is normally computed via backend.
This ensures that if you have different conf by env
you won't mess up w/ images coming from other envs (eg: prod).

This change makes sure that the URL which is used for the thumbs
is always computed by the backend.

### DIFF
--- a/storage_image/models/storage_image.py
+++ b/storage_image/models/storage_image.py
@@ -47,6 +47,7 @@ class StorageImage(models.Model):
             vals["backend_id"] = self._get_default_backend_id()
         # When using the widget image_url, the create will pass the data
         # in the "url" field. We map it to the data field
+        # TODO: WHY??? I don't see any handling of `data` as url in storage.file
         for key in ["image_medium_url", "image_small_url"]:
             if key in vals:
                 vals["data"] = vals.pop(key)
--- a/storage_image_product/models/product_product.py
+++ b/storage_image_product/models/product_product.py
@@ -24,8 +24,12 @@ class ProductProduct(models.Model):
     )
     # small and medium image are here to replace
     # native image field on form and kanban
-    variant_image_small_url = fields.Char(related="main_image_id.image_small_url")
-    variant_image_medium_url = fields.Char(related="main_image_id.image_medium_url")
+    variant_image_small_url = fields.Char(
+        string="Variant main small image URL", related="main_image_id.image_small_url"
+    )
+    variant_image_medium_url = fields.Char(
+        string="Variant main medium image URL", related="main_image_id.image_medium_url"
+    )
 
     @api.depends(
         "product_tmpl_id.image_ids.attribute_value_ids",
--- a/storage_image_product/models/product_product.py
+++ b/storage_image_product/models/product_product.py
@@ -1,5 +1,7 @@
 # Copyright 2017 Akretion (http://www.akretion.com).
 # @author Sébastien BEAU <sebastien.beau@akretion.com>
+# Copyright 2021 Camptocamp SA (http://www.camptocamp.com)
+# @author Simone Orsi <simahawk@gmail.com>
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 
 from odoo import api, fields, models
@@ -8,24 +10,22 @@ from odoo import api, fields, models
 class ProductProduct(models.Model):
     _inherit = "product.product"
 
-    # small and medium image are here to replace
-    # native image field on form and kanban
-    variant_image_small_url = fields.Char(
-        compute="_compute_variant_image_ids",
-        store=True,
-        string="Variant Image Small Url",
-    )
-    variant_image_medium_url = fields.Char(
-        compute="_compute_variant_image_ids",
-        store=True,
-        string="Variant Image Medium Url",
-    )
     variant_image_ids = fields.Many2many(
         "product.image.relation",
         compute="_compute_variant_image_ids",
         store=True,
         string="Variant Images",
     )
+    main_image_id = fields.Many2one(
+        "storage.image",
+        compute="_compute_main_image_id",
+        # Store it to improve perf on product views
+        store=True,
+    )
+    # small and medium image are here to replace
+    # native image field on form and kanban
+    variant_image_small_url = fields.Char(related="main_image_id.image_small_url")
+    variant_image_medium_url = fields.Char(related="main_image_id.image_medium_url")
 
     @api.depends(
         "product_tmpl_id.image_ids.attribute_value_ids",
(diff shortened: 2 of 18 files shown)


simahawk:
@sebastienbeau @lmignon @bguillot what was the reason for storing images' url?
Can you check my `TODO`s in this PR?

storage_image/models/storage_image.py
```diff
@@ -47,6 +47,7 @@ def create(self, vals):
             vals["backend_id"] = self._get_default_backend_id()
         # When using the widget image_url, the create will pass the data
         # in the "url" field. We map it to the data field
+        # TODO: WHY??? I don't see any handling of `data` as url in storage.file
```
simahawk, line 50: Would be nice to know why this was needed :)

simahawk:
@sebastienbeau @lmignon @bguillot or @rousseldenis anyone? :pray:

storage_image_product/models/product_product.py
```diff
@@ -11,6 +11,7 @@ class ProductProduct(models.Model):
     # small and medium image are here to replace
     # native image field on form and kanban
     variant_image_small_url = fields.Char(
+        # TODO: why store?
```
lmignon: store=True is probably required to avoid perf issues :thinking:

lmignon (approved):
LGTM (code review only)

storage_image_product/models/product_product.py
simahawk: @lmignon yeah but at this stage the thumbs are already there and the computation should not take that much.
Maybe, it has a big impact on the kanban view?
BUT storing the URL is really a pain when you restore databases across different envs.

Ahhh, while thinking again on this (right now) I think we should store the relation to the thumbnails instead of the URL.
This way the computation should should be much faster. What do you think?

simahawk:
@lmignon could you pass by again? I've refactored product image url computation in the commit `storage_image_product: do not store images URLs`.

CC @JordiBForgeFlow @HviorForgeFlow

MiquelRForgeFlow:
@simahawk could you squash some commits? AFAIK, it should only be 2 commits: the refactor of tests and the non storing urls... This will give a clear understanding of the overall changes.

storage_image_product/models/product_product.py
```diff
@@ -8,24 +10,22 @@
...
     )
+    main_image_id = fields.Many2one(
+        "storage.image",
+        compute="_compute_main_image_id",
+        # Store it to improve perf on product views
+        store=True,
+    )
+    # small and medium image are here to replace
+    # native image field on form and kanban
+    variant_image_small_url = fields.Char(related="main_image_id.image_small_url")
```
lmignon: @simahawk IMO it's important to keep the image urls stored on the product template and on the variant to avoid extra queries when the product kanban view is displayed. 
If you change your env, you are not sure at all that your images exist into the other env. In this case, the new computed url could not exist.

simahawk:
> @simahawk could you squash some commits? AFAIK, it should only be 2 commits: the refactor of tests and the non storing urls... This will give a clear understanding of the overall changes.

hmm that's the contrary in theory: all my commits are atomic and you can see from each commit what I changed exactly.

storage_image_product/models/product_product.py
simahawk: @lmignon but THIS is the problem. If you change environment you should not see the images IF they are not there. Otherwise, what happens is that you expect to have those images in the system but you don't. Then the customer or a functional colleague comes to you and asks "hey, I see my images in Odoo, why I don't see them in the shop (or any other service)?".

simahawk: also. if you need to change CDN URL on production, it won't have any effect on those URLs. That sucks IMO. One of the key point about storage files is that their URL is controlled by the backend.

lmignon: > @lmignon but THIS is the problem. If you change environment you should not see the images IF they are not there. Otherwise, what happens is that you expect to have those images in the system but you don't. Then the customer or a functional colleague comes to you and asks "hey, I see my images in Odoo, why I don't see them in the shop (or any other service)?".

If you change environment and you don't want to see the image, you should apply a script to remove   `product.image.relation` and related `storage.image`and to set urls to blank.
Sometime it's nice to be able to see the images from an other env. (ex test env can see image from prod env but is not allowed to modify these items.)

lmignon: > also. if you need to change CDN URL on production, it won't have any effect on those URLs. That sucks IMO. One of the key point about storage files is that their URL is controlled by the backend.

I prefer to write a script to upgrade urls by hand than fighting against performance issue... :smirk:

We could also add a listener on the write of `storage.backend` to update image urls if the URL changes on the backend in a efficient way (SQL ...)

simahawk: > Sometime it's nice to be able to see the images from an other env. (ex test env can see image from prod env but is not allowed to modify these items.)

That's another kind of problem. We can work on it. Also, setting the production CDN on other envs won't have any impact on what you can write -> easy to do, fast result, no script / black magic needed.

> I prefer to write a script to upgrade urls by hand than fighting against performance issue... smirk

I think you are not considering the changes here: very likely perf were bad because of full awakening on several m2m fields: 

* rel w/ images
* rel w/ attribute values from the image relation
* rel w/ product attribute values on variants

a lot of m2m loaded every single time which we know to be very slow!
With this change you have only a m2o to look for.

Furthermore, the URL before my refactoring on https://github.com/OCA/storage/pull/88 was fully computed. The most expensive thing being the slug, which is now stored.

And now that I think about this again I think I'm going to store the main thumbs relation as well because to compute the thumb URL we always loop on all the thumbs and it can be expensive.

simahawk: @lmignon so, in [this commit](https://github.com/OCA/storage/pull/94/commits/596dcf7a73fd4da9fd4798c4ef964a0f1350e4af)  `fixup! storage_thumb: do not store url` I've made the main thumbs computed. This ensures that the URL computation is very fast. I've tried on a kanban w/ thousands of records.

I've done a test by playing around w/ storage backend server env params and - as expected - if you keep the CDN from production (or whatever env the db is coming from) you'll see the images w/out any problem.

The only issue that I've found is that `base_url_for_files` is not refreshed when changing server env params as the change is not detected. Hence, I added a btn to refresh the base URL on the fly (this affects only the backend, no storage file computation is involved).

Finally, I made the 2 thumbs fields visible and editable so if by chance you want to take control of which image is used as thumbs, you can do it. No additional cost here.

Sounds good?

lmignon (approved):
Thank you for all your hard work @simahawk  You enhancements should improve the UX without introducing side effects. 
LGTM (Code review only) 
Unfortunately travis is :red_circle:

simahawk:
build is :green_circle: 

@MiquelRForgeFlow mind to give a review?

MiquelRForgeFlow:
sorry, I won't review. hope @HviorForgeFlow has some time :)

PR #96, opened 2021-05-03
[12.0][FIX] storage_image: more robust security rule + FIX 12.0 repository configuration

len-foss:
The fact that this rules does not apply in read means
that a create could have laxer restrictions than a read;
meaning that a create could actually fail on an access rights error.
This is actually the case when installing for example storage media.


rousseldenis (approved):

rousseldenis:
@len-foss Could you check cryptography warning on runbot ?

len-foss:
@rousseldenis Upgraded Python to 3.6 in last commit, that seems like the most sensible thing.

PR #97, opened 2021-05-10
[14][storage_backend] filter files in _list_files by pattern

florian-dacosta:
Hello,

This feature was added in version 12, but a PR with migration was already ongoing, so it has been lost.
Re-introduce the feature, coming from https://github.com/OCA/storage/pull/35 and needed for https://github.com/OCA/server-tools/pull/2001

@kevinkhao @sebastienbeau @lmignon


kevinkhao (approved):
LGTM

sebastienbeau (approved):

PR #99, opened 2021-06-04
[13.0][FIX] storage_image: delete relation when image is deleted

mmequignon:
Deleting an image from the storage image menu doesn't delete the relation, and the image is still referenced.

### DIFF
--- a/storage_image/models/storage_relation_abstract.py
+++ b/storage_image/models/storage_relation_abstract.py
@@ -16,7 +16,7 @@ class ImageRelationAbstract(models.AbstractModel):
     _order = "sequence, image_id"
 
     sequence = fields.Integer()
-    image_id = fields.Many2one("storage.image", required=True)
+    image_id = fields.Many2one("storage.image", required=True, ondelete="cascade")
     # for kanban view
     image_name = fields.Char(related="image_id.name")
     # for kanban view


TDu (approved):

storage_image/models/storage_relation_abstract.py
```diff
@@ -16,7 +16,7 @@ class ImageRelationAbstract(models.AbstractModel):
     _order = "sequence, image_id"
 
     sequence = fields.Integer()
-    image_id = fields.Many2one("storage.image", required=True)
+    image_id = fields.Many2one("storage.image", required=True, ondelete="cascade")
```
simahawk, line 19: I'm not sure this is a good idea every time. Maybe we want this only from the product form? 
All in all you could potentially use an image for many records. Probably this `ondelete` has been skipped on purpose.
I would override `unlink` and use a ctx key to control the behavior.
Then in the product form, set this ctx key. Or, it can be configuration on the storage.
Alternatively you could use `restrict` and show a btn to manage images for the current record and from there you could delete the images 1st.

I know it sounds complicated but maybe users won't expect to delete images right away.
@lmignon @sebastienbeau what's your POV?

lmignon: I agree with @simahawk ...

>  Alternatively you could use restrict and show a btn to manage images for the current record and from there you could delete the images 1st.

:+1: 

An other motivation to not `unlink` an image is to avoid dead link into html pages where the image is referenced..

ivantodorovich: Note that this is not really deleting the image, it's deleting the `image.relation` record which links the image with other records like products, categories, etc..

The `image.relation` is a model that could be thought as a many2many relation table, only that you can add fields to the relation itself. It doesn't make sense to keep its records if any of the two records in the relation are removed.

I've implemented the same (for 14.0) in: https://github.com/OCA/storage/pull/103

ivantodorovich (approved):

storage_image/models/storage_relation_abstract.py
simahawk: You are right, I read it the other way around indeed :sweat:

simahawk (approved):

storage_image/models/storage_relation_abstract.py
simahawk: @lmignon ping

lmignon: Thank you for the extra informations @ivantodorovich. Indeed, I was wrong on the potential effect of such a cascade delete

lmignon (approved):

PR #101, opened 2021-06-25
[13.0][FIX] storage_media: Explicit ir rule definition for Storage File for Media

HviorForgeFlow:
If Apply for Read is kept it collides with Read for Public Users when storage backend is public.

CC @ForgeFlow


simahawk (approved):

sebastienbeau (approved):
LGTM (code review)
Thanks for your contribution

PR #105, opened 2021-07-30
[14.0] storage_file: boost url compute

simahawk:
FWD port of https://github.com/OCA/storage/pull/88

### DIFF
--- a/storage_file/models/ir_actions_report.py
+++ b/storage_file/models/ir_actions_report.py
@@ -1,0 +1,14 @@
+# Copyright 2021 Camptocamp SA (http://www.camptocamp.com).
+# @author Simone Orsi <simahawk@gmail.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class IrActionsReport(models.Model):
+    _inherit = "ir.actions.report"
+
+    def render_qweb_pdf(self, res_ids=None, data=None):
+        return super(
+            IrActionsReport, self.with_context(print_report_pdf=True)
+        ).render_qweb_pdf(res_ids=res_ids, data=data)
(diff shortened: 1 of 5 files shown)


mmequignon (approved):

storage_file/models/ir_actions_report.py
```diff
@@ -0,0 +1,14 @@
...
+from odoo import models
+
+
+class IrActionsReport(models.Model):
+    _inherit = "ir.actions.report"
+
+    def render_qweb_pdf(self, res_ids=None, data=None):
+        return super(
+            IrActionsReport, self.with_context(print_report_pdf=True)
+        ).render_qweb_pdf(res_ids=res_ids, data=data)
```
ivantodorovich, line 14:
```suggestion
        self = self.with_context(print_report_pdf=True)
        return super().render_qweb_pdf(res_ids=res_ids, data=data)
```
Not sure what you think but I recently discovered this is the same and it's more readable.
Not blocking of course

storage_file/models/storage_backend.py
```diff
@@ -79,3 +80,42 @@ def _get_backend_id_from_param(cls, env, param_name, default_fallback=True):
...
+        "directory_path",
+        "url_include_directory_path",
+    )
+    def _compute_base_url_for_files(self):
+        for record in self:
+            record.base_url_for_files = record._get_base_url_for_files()
+
+    def _get_base_url_for_files(self):
+        """Retrieve base URL for files."""
+        backend = self.sudo()
```
ivantodorovich, line 96:
```suggestion
```

Is this necessary?
In this method compute is stored so it already runs on sudo, for the other cases maybe consider `compute_sudo=True`

ivantodorovich (approved):
not blocking of course
LG :+1:

ivantodorovich:
@simahawk can we merge this?

PR #110, opened 2021-08-17
[13.0] storage_file: add backend display policy

simahawk:
Main goals:

1. avoid loading external binary data when not necesary
2. avoid crash if the external storage is not reachable but its CDN is available

Additional change: do not call `sudo` for each record.


storage_file/README.rst
```diff
@@ -33,6 +33,7 @@ Contributors
 
 * Sebastien Beau <sebastien.beau@akretion.com>
 * Raphaël Reverdy <raphael.reverdy@akretion.com>
+* Simone Orsi <simone.orsi@camptocamp.com>
```
simahawk: this is actually due to all my past contribs :sweat_smile:

storage_file/__manifest__.py
```diff
@@ -9,6 +9,7 @@
     "category": "Storage",
     "website": "https://github.com/OCA/storage",
     "author": " Akretion, Odoo Community Association (OCA)",
+    "maintainers": ["sebastienbeau", "hparfr", "simahawk"],
```
simahawk, line 12: @sebastienbeau @hparfr you don't mind right? :smile:

sebastienbeau: Ok for us ;)

storage_file/views/storage_file_view.xml
```diff
@@ -27,9 +27,11 @@
                             name="backend_id"
                             attrs="{'readonly': [('id', '!=', False)]}"
                         />
+                        <field name="is_external" />
                         <field
-                            name="data"
-                            attrs="{'readonly': [('id', '!=', False)]}"
+                            name="data_display"
```
simahawk, line 32: hmm this change prevents editing of the file. I'll find a better solution.

simahawk:
note to myself: rework on top of https://github.com/OCA/storage/pull/154

PR #100, opened 2021-06-25
[ADD] 14.0_storage_image_import

Kev-Roche:

PR #113, opened 2021-08-26
[13.0][ADD] storage_import_image_advanced

HviorForgeFlow:
Accomodation of [shopinvader_import_image](https://github.com/shopinvader/odoo-shopinvader/tree/13.0/shopinvader_import_image) into OCA/storage repository

CC @ForgeFlow @simahawk


simahawk:
You read my mind... I was planning this since a while :smile: 
Can you cherry-pick these changes?  https://github.com/shopinvader/odoo-shopinvader/pull/1077

HviorForgeFlow:
> You read my mind... I was planning this since a while 
> Can you cherry-pick these changes? [shopinvader/odoo-shopinvader#1077](https://github.com/shopinvader/odoo-shopinvader/pull/1077)

Done!

.travis.yml
```diff
@@ -12,6 +12,7 @@ addons:
   apt:
     packages:
       - expect-dev # provides unbuffer utility
+      - libmagic1 # Required by module storage_import_image
```
simahawk, line 15: I think this should be added to .copier-answers.yml

simahawk (approved):

.travis.yml
HviorForgeFlow: not sure how to add this dependency there :face_with_head_bandage:

simahawk: the issue is that AFAIR we'll lose the dep while sync'ing w/ https://github.com/OCA/oca-addons-repo-template.
I think it's just fine to add it here https://github.com/OCA/storage/blob/14.0/.copier-answers.yml#L12
@Yajo any recommendation in such cases?

yajo: Copier update will respect your git history, so as long as the answers-vs-code relation don't change, you won't get conflicts. However indeed it's better IMHO to just do `copier update` and add that new dependency.

sebastienbeau:
Hi we have an alternative module already merge in V14 for import image see PR (merged) here https://github.com/OCA/storage/pull/100

With this module you have a field "import_from_url" on the image relation and you can use native odoo import (or xmlrpc or whatever you want) to field the field so it will import the image.

I do not know if we should drop the shopinvader_image_import module or found a name that is quite different with the name of module in V14

sebastienbeau:
Note in case that you want to import xls or csv file using chunk and job there is also this generic module : https://github.com/shopinvader/pattern-import-export

simahawk:
> Hi we have an alternative module already merge in V14 for import image see PR (merged) here #100
> 
> With this module you have a field "import_from_url" on the image relation and you can use native odoo import (or xmlrpc or whatever you want) to field the field so it will import the image.

@sebastienbeau I think that for most use cases that approach is quite limited as you don't have configuration of storages, no jobs splitting, no state. I think we should simply mentioned the 2 modules in each readme and that's it. 
I don't know if that was your point... ? :smile: 
 
> I do not know if we should drop the shopinvader_image_import module or found a name that is quite different with the name of module in V14

IMO we should deprecate it for 13 and 14 and trash it on 15.

sebastienbeau:
It's fine for me to have the both module (customer will decide if both should survive in long term).
I was just saying that maybe we should have a name that is more different.
Indeed in V14 we have a module called "storage_image_import" so having a second module with the name "storage_import_image" is quite dangerous.
Maybe something like storage_image_csv_import  will be better ?

simahawk:
I see. Yep, you are right. I'd go for `storage_image_import_advanced`. What do you think?

sebastienbeau:
Ok for me

simahawk:
@sebastienbeau good for you?

sebastienbeau (approved):
Ok for me

simahawk:
/ocabot merge patch


@HviorForgeFlow do you plan fwd port to 14?

simahawk:
meh, was already merged but I had no updated UI.... new build canceled :sweat_smile:

PR #114, opened 2021-08-30
[14.0][IMP] add tag support for brand image

sebastienbeau:
@ivantodorovich to be consistent with product and category

### DIFF
--- a/storage_image_product_brand/views/product_brand_image_relation.xml
+++ b/storage_image_product_brand/views/product_brand_image_relation.xml
@@ -14,6 +14,12 @@
             <form position="attributes">
                 <attribute name="string">Images</attribute>
             </form>
+            <group name="extra" position="inside">
+                <field
+                    name="tag_id"
+                    context="{'active_model': context.get('active_model')}"
+                />
+            </group>
         </field>
     </record>
 
--- a/storage_image_product_brand/models/__init__.py
+++ b/storage_image_product_brand/models/__init__.py
@@ -1,2 +1,3 @@
 from . import product_brand
 from . import product_brand_image_relation
+from . import image_tag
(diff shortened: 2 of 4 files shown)


storage_image_product_brand/views/product_brand_image_relation.xml
```diff
@@ -14,6 +14,12 @@
             <form position="attributes">
                 <attribute name="string">Images</attribute>
             </form>
+            <group name="extra" position="inside">
+                <field
+                    name="tag_id"
+                    context="{'active_model': context.get('active_model')}"
```
ivantodorovich, line 20: When I saw the same code in the other modules i wondered: is this really necessary?
If `active_model` is already in context, why re-setting it?

ivantodorovich (approved):

storage_image_product_brand/views/product_brand_image_relation.xml
sebastienbeau: In fact the field tag_id is doesn't not exist in the abstract model and not in the abstract view.
Maybe we can add it by default ?

sebastienbeau: I just realise that I didn't understand you question.
Yes it's needed, indeed odoo do not propagate the context by default for the many2one during the quick create, so we need to pass it explicitly. Just tested right now

ivantodorovich: Good to know 👍 thanks!

simahawk: good to know, indeed :smile: :pray:

simahawk (approved):

PR #115, opened 2021-09-01
[13.0][IMP]storage_image_product: main image of the variant

AaronHForgeFlow:
@ForgeFlow @HviorForgeFlow

### DIFF
--- a/storage_image_product/tests/test_product_image_relation.py
+++ b/storage_image_product/tests/test_product_image_relation.py
@@ -114,3 +114,39 @@ class ProductImageCase(ProductImageCommonCase):
             (self.logo_image, self.product_b),
         )
         self._test_main_images_and_urls(expected)
+
+    def test_main_image_attribute(self):
+        """
+        Attach the image to the template and check the first image of the
+        variant is the one with same attributes
+        """
+        self.env["product.image.relation"].create(
+            {
+                "product_tmpl_id": self.template.id,
+                "image_id": self.logo_image.id,
+                "sequence": 1,
+            }
+        )
+        self.env["product.image.relation"].create(
+            {
+                "product_tmpl_id": self.template.id,
+                "image_id": self.white_image.id,
+                "attribute_value_ids": [
+                    (
+                        6,
+                        0,
+                        [
+                            self.env.ref("product.product_attribute_value_4").id,
+                            self.env.ref("product.product_attribute_value_1").id,
+                        ],
+                    )
+                ],
+                "sequence": 10,
+            }
+        )
+        # The variant should not take the only with the lowest sequence but
+        # the one with same attributes
+        expected = ((self.white_image, self.product_b),)
+        self._test_main_images_and_urls(expected)
+        expected = ((self.logo_image, self.product_c + self.product_a),)
+        self._test_main_images_and_urls(expected)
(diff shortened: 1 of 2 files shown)


simahawk:
this is similar to https://github.com/OCA/storage/pull/116 but for images right?

AaronHForgeFlow:
When the images are attached to the template and not to the variant. In that case in the variant it shows an image without attributes. The fix search first for images with the same attributes than the variant. 

Perhaps this is not a common use case as long I am storing all the images in the template instead of storing them in any image individually. If that is the case I will close this.

sebastienbeau:
Hi
I think I understand.
If you put 3 images on a product template in this order
- a generic image for all product (like a image of the situtation, or an image with all the variant visible)
- a specific image for the blue product
- a specific image for the red product

With the existing implementation, the generic one will used as main image for the product template and for all variant.
With your change each variant will try to show the first "specific image" that match for them. So in our case it will a be the blue image for the blue product, the red image for the red product.

I am correct ?

If yes it's seem to be a good idea, can you add a test? If needed i can do it

Thanks

AaronHForgeFlow:
> If you put 3 images on a product template in this order
> 
> a generic image for all product (like a image of the situtation, or an image with all the variant visible)
> a specific image for the blue product
> a specific image for the red product
> With the existing implementation, the generic one will used as main image for the product template and for all variant.
> With your change each variant will try to show the first "specific image" that match for them. So in our case it will a be the blue image for the blue product, the red image for the red product.

Yes, exactly :)

I will try to add a test Thanks for the review!

AaronHForgeFlow:
Test added :)

storage_image_product/tests/test_product_image_relation.py
```diff
@@ -114,3 +114,39 @@ def test_main_image_and_urls(self):
             (self.logo_image, self.product_b),
         )
         self._test_main_images_and_urls(expected)
+
+    def test_main_image_attribute(self):
+        """
+        Attach the image to the template and check the first image of the
+        variant is the one with same attriubttes
```
gurneyalex:
```suggestion
        variant is the one with same attributes
```

storage_image_product/tests/test_product_image_relation.py
```diff
@@ -114,3 +114,39 @@ def test_main_image_and_urls(self):
...
+                        [
+                            self.env.ref("product.product_attribute_value_4").id,
+                            self.env.ref("product.product_attribute_value_1").id,
+                        ],
+                    )
+                ],
+                "sequence": 10,
+            }
+        )
+        # The variant should not take the only wiht lower sequencce but
```
gurneyalex:
```suggestion
        # The variant should not take the only with the lowest sequence but
```

gurneyalex (approved):

simahawk (approved):

sebastienbeau:
@AaronHForgeFlow can you add the change propose by alex ? so we can merge this ? Thansk

AaronHForgeFlow:
> @AaronHForgeFlow can you add the change propose by alex ? so we can merge this ? Thansk

Done! Thanks! :)

sebastienbeau:
Thanks @AaronHForgeFlow for your work !

PR #123, opened 2021-10-01
[13.0][FIX] storage_image: Allow read storage file permission on image manager

HviorForgeFlow:
CC @ForgeFlow


JordiBForgeFlow (approved):

sebastienbeau:
Hi can you give me an example of use case?
Indeed if I look at this PR here https://github.com/OCA/storage/pull/101/files we do not apply the ir rule for the read.
So it's quite the same case.
I am not sure that we should apply this restrcition

PR #129, opened 2021-10-06
[IMP] storage_backend: Avoid FileNotFoundError when deleting files on filestore

HviorForgeFlow:
It should not happen, but, why not if file is not on the system just continue with the cleanup

CC @ForgeFlow


sebastienbeau (approved):
Thanks for the improvement

hparfr (approved):

GSLabIt:
is this ready to merge?

gurneyalex (approved):

PR #131, opened 2021-10-06
[FIX] storage_image_product_brand: fix dependency on storage_image_product instead of storage_image

kevinkhao:


kevinkhao:
in this module we modify image.tag which is declared in storage_image_product

simahawk (approved):

sebastienbeau (approved):

PR #149, opened 2021-12-08
[14][MIG] storage_backend_ftp

florian-dacosta:
Standard Migration

I have tested sending a simple file on a real FTP server, and it worked properly.

@LoisRForgeFlow @acsonefho


florian-dacosta:
@simahawk I am not sure about the pre-commit error, do you know what is it?
Should I update the requirement.txt with the missing lib in the PR?

simahawk:
> @simahawk I am not sure about the pre-commit error, do you know what is it? Should I update the requirement.txt with the missing lib in the PR?

you just have to run pre-commit locally and add requirements.txt to staged files -> the dependency is missing

florian-dacosta:
Ok done, thanks Simone.
I thought it was done automatically somehow, like the readme file.

sebastienbeau (approved):
LGTM code review

LoisRForgeFlow:
Hi @florian-dacosta 

We have found a problem using the FTP over TLS connection of this module in v13. We haven't used in the past that option until now, when we found that it was not working as expected. We created this PR fixing the functionality https://github.com/OCA/storage/pull/153.

My questions comes now, do you use the current TLS option? does it work for you? I want to make sure that it was actually a malfunctioning code.

florian-dacosta:
Hi @LoisRForgeFlow 
I don't use the module in production yet (and won't before a few month I think)
I just made a quick test after the port of the module with a simple FTP transfer (no TLS) and that went fine.
In previous versions of Odoo I never needed FTP over TLS actually, it  was simple FTP or SFTP.

PR #153, opened 2021-12-14
[13.0] [IMP] storage_backend_ftp

JasminSForgeFlow:
Fix connection problem for implicit FTP over TLS
@ForgeFlow

### DIFF
--- a/storage_backend_ftp/components/ftp_adapter.py
+++ b/storage_backend_ftp/components/ftp_adapter.py
@@ -41,31 +41,46 @@ def ftp_mkdirs(client, path):
             raise  # pragma: no cover
 
 
+class ImplicitFTPTLS(ftplib.FTP_TLS):
+    """FTP_TLS subclass that automatically wraps sockets in SSL to support implicit FTPS."""
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._sock = None
+
+    @property
+    def sock(self):
+        """Return the socket."""
+        return self._sock
+
+    @sock.setter
+    def sock(self, value):
+        """When modifying the socket, ensure that it is ssl wrapped."""
+        if value is not None and not isinstance(value, ssl.SSLSocket):
+            value = self.context.wrap_socket(value)
+        self._sock = value
+
+
 @contextmanager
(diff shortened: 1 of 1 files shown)


LoisRForgeFlow:
Hi @acsonefho 

We are modifying your original code to connect over TLS, it seemed to cover an explicit FTP over TLS connection. You included a comment in the code:

`# Due to a bug into between ftplib and ssl, this part (about ssl) might not work!`

That makes me think that you haven't used the connection over TLS functionality, can you confirm?. Assuming so, we have changed the approach to  use implicit FTP over TLS. 

Is this change fine with you? or maybe it was really working for you, so in such case we should add a new ftp_encryption option to have FTP, implicit FTPS and explicit FTPS to cover all 3 cases.

What do you think?

acsonefho:
> Hi @acsonefho
> 
> We are modifying your original code to connect over TLS, it seemed to cover an explicit FTP over TLS connection. You included a comment in the code:
> 
> `# Due to a bug into between ftplib and ssl, this part (about ssl) might not work!`
> 
> That makes me think that you haven't used the connection over TLS functionality, can you confirm?. Assuming so, we have changed the approach to use implicit FTP over TLS.
> 
> Is this change fine with you? or maybe it was really working for you, so in such case we should add a new ftp_encryption option to have FTP, implicit FTPS and explicit FTPS to cover all 3 cases.
> 
> What do you think?

Feel free to edit what is necessary :)

LoisRForgeFlow (approved):
Functional review, works as expected :+1:

LoisRForgeFlow:
@OCA/storage-maintainers can you check this? It seems that the author of the module in v13 and the contributor migrating it to v14 did not really used FTPS, so I guess this PR is good to go.

LoisRForgeFlow:
@simahawk can you help getting this merge? Thanks!

simahawk (approved):
LG

simahawk:
/ocabot merge minor

To be fwd ported I guess?

LoisRForgeFlow:
> /ocabot merge minor
> 
> To be fwd ported I guess?

Indeed, @JasminSForgeFlow could you forward port this?

PR #154, opened 2021-12-17
storage_thumbnail: use internal URL on demand

simahawk:
Using the external URL for internal usage (eg: backend views)
generates useless traffic over CDN which might be costly.

Now you can configure on the backend if you want to use the internal url or the public one.

### DIFF
--- a/storage_file/models/storage_file.py
+++ b/storage_file/models/storage_file.py
@@ -34,6 +34,10 @@ class StorageFile(models.Model):
         "storage.backend", "Storage", index=True, required=True
     )
     url = fields.Char(compute="_compute_url", help="HTTP accessible path to the file")
+    internal_url = fields.Char(
+        compute="_compute_internal_url",
+        help="HTTP URL to load the file directly from storage.",
+    )
     slug = fields.Char(
         compute="_compute_slug", help="Slug-ified name with ID for URL", store=True
     )
@@ -88,9 +92,9 @@ class StorageFile(models.Model):
             record.slug = record._slugify_name_with_id()
 
     def _slugify_name_with_id(self):
-        return u"{}{}".format(
+        return "{}{}".format(
             slugify(
-                u"{}-{}".format(self.filename, self.id), regex_pattern=REGEX_SLUGIFY
+                "{}-{}".format(self.filename, self.id), regex_pattern=REGEX_SLUGIFY
             ),
             self.extension,
         )
@@ -148,6 +152,19 @@ class StorageFile(models.Model):
     def _get_url(self):
         """Retrieve file URL based on backend params."""
         return self.backend_id._get_url_for_file(self)
+
+    @api.depends("slug")
+    def _compute_internal_url(self):
+        for record in self:
+            record.internal_url = record._get_internal_url()
+
+    def _get_internal_url(self):
+        """Retrieve file URL to load file directly from the storage.
+
+        It is recommended to use this for Odoo backend internal usage
+        to not generate traffic on external services.
+        """
+        return f"/storage.file/{self.slug}"
 
     @api.depends("name")
     def _compute_extract_filename(self):
--- a/storage_thumbnail/models/thumbnail_mixin.py
+++ b/storage_thumbnail/models/thumbnail_mixin.py
@@ -28,20 +28,20 @@ class ThumbnailMixing(models.AbstractModel):
         compute="_compute_main_thumbs",
         store=True,
         readonly=False,
+        compute_sudo=True,
     )
     thumb_small_id = fields.Many2one(
         comodel_name="storage.thumbnail",
         compute="_compute_main_thumbs",
         store=True,
         readonly=False,
+        compute_sudo=True,
     )
     image_medium_url = fields.Char(
-        string="Medium thumb URL",
-        compute="_compute_thumb_urls",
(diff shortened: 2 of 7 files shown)


simahawk:
@lmignon @sebastienbeau could you provide your feedback? I see no reason to load images from CDN on backend views.

i-vyshnevska (approved):

SilvioC2C (approved):

storage_thumbnail/models/thumbnail_mixin.py
```diff
@@ -54,6 +56,17 @@ def _compute_main_thumbs(self):
                 fname = "thumb_%s_id" % scale
                 rec[fname] = rec._get_thumb(scale_key=scale)
 
+    @api.depends_context("storage_public_url")
+    @api.depends("thumb_medium_id", "thumb_small_id")
+    def _compute_thumb_urls(self):
+        # Use always the internal URL when not required otherwise.
+        url_fname = "internal_url"
+        if self.env.context.get("storage_public_url"):
```
lmignon: PLZ no parameter through context....

IMO internal url is only a good idea when images are stored by odoo. It should therefore  not be the default.

simahawk: > PLZ no parameter through context....

That's the only way to gain control from specific points (eg: related fields). 
Also, I don't see why this should be bad for such a case.

> IMO internal url is only a good idea when images are stored by odoo. It should therefore not be the default.

This is exactly the contrary of what we are experiencing.
CDN traffic burnt for internal ERP usage -> are you sure is a good idea?
CDNs are supposed to be used for large public.

The change here affects only the URLs of the images showed in product views.

lmignon: > The change here affects only the URLs of the images showed in product views.

It"s on purpose it's done as it's. If these images are available on a CDN we don't want to burn odoo's resources to serve theses binaries contents.

simahawk: Now I'm curious about which service do you use because it sounds very cheap (if not free) :smile: 
Anyway, I'll adapt this to become an explicit setting in the backend.

lmignon: You could put your files on a server via ftp and serve theses files via a reverse proxy on a dedicated url, you could use an S3 compatible server as https://min.io/ ...

simahawk: I think you are mixing up "CDN" and "Object Storage".

Reading files publicly from the object storage is _not_ recommended. You need CDN in front of it for public resources.

We do use min.io too (we'll migrate our infra to it actually) and when you add a CDN in front is not for free.
AFAIK every CDN (from AWS or Azure or whatever) is going to cost you based on traffic.

That being said, I'll see if we have alternatives. Having a specific reverse proxy in front of the storage adds another layer to maintain in the stack.

But then, we'll have to enhance the configuration of the backend to be able to configure a CDN URL for public usage and another CDN-like URL for internal usage. ATM you can set only one URL for all resources.

lmignon: @simahawk I thing that something exists to rewrite the url when image is exported for a specific website into shopinvader. https://github.com/shopinvader/odoo-shopinvader/tree/13.0/shopinvader_backend_image_proxy

simahawk: Thanks for the pointer :)

Nevertheless, seems weird to me that the "public URL" of the file is not the one that you configure on the backend.

Also, such feature, the proxy - or the distinction that I make here between public and internal usage URLs - should stay here in the storage repo IMO and computed on a specific field.Then w/ a policy setting on the backend you can decide what to use and when. 

I would keep the changes in this PR and:

1. add a flag `internal_url_for_backend_views` to enable the behavior. When OFF, it behaves as today.
2. add a field `internal_base_url`, default= `/storage.file/$slug`, to give control on internal URLs 

This change would make `s_backend_image_proxy` obsolete because `.url` will always be the public one (normally using a CDN).

What do you think?

simahawk:
@lmignon I refactored and added a flag `backend_view_use_internal_url` on the storage backend to control this behavior.
Could you have a look again?

storage_file/models/storage_file.py
```diff
@@ -149,6 +153,19 @@ def _get_url(self):
         """Retrieve file URL based on backend params."""
         return self.backend_id._get_url_for_file(self)
 
+    @api.depends("slug")
+    def _compute_internal_url(self):
+        for record in self:
+            record.internal_url = record._get_internal_url()
+
+    def _get_internal_url(self):
```
lmignon, line 161: Why do you need an other method than the compute one?

lmignon (approved):
IMO the `_get_internal_url` method is useless. Apart from this LGTM (Code review)

storage_file/models/storage_file.py
simahawk: I tend to split out the computation for single records most of the times because I hate to have too loop twice on the same recordset (by calling super) when you want to change the behavior of a computed field.
As it does not hurt, I'd like to keep it :wink:

simahawk:
BTW it would be nice to start using GH actions but https://github.com/OCA/storage/pull/184 is blocked by https://github.com/OCA/brand/pull/118

PR #155, opened 2021-12-20
s_backend_s3: refactor get bucket

simahawk:
* avoid calling s3.meta.client.head_bucket on each get, because it's slow
* rationalise implementation

TODO: tests

### DIFF
--- a/storage_backend_s3/components/s3_adapter.py
+++ b/storage_backend_s3/components/s3_adapter.py
@@ -21,6 +21,10 @@ except ImportError as err:  # pragma: no cover
     _logger.debug(err)
 
 
+# Keep track of existing buckets and avoid checking for them every time.
+EXISTING_BUCKETS = []
+
+
 class S3StorageAdapter(Component):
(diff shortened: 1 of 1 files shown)


simahawk:
After thinking again about this I'd go for:
1. remove global var
2. add flag `s3_bucket_exists` on the backend
3. btn `action_ensure_bucket_exists` to test if the bucket exists by calling `_ensure_bucket_exists`
4. modify `_check_bucket_exists` to accept a flag `force=False` 
5. if the force flag is false and `self.s3_bucket_exists` is True return True 
6. if test is ok, set the flag
7. if the bucket is not available, create it and set the flag
8. no more call to check if the bucket exists

Rationale: in real life we check if the bucket exists only once when preparing the backend.
If the bucket suddenly disappears you will see bad errors anyway, there's no need to check forever or every time a worker is started. 
If you really want to check it more often you can create a cron that checks it from time to time and send a msg if is not, but this is not required here.

PR #157, opened 2021-12-23
storage_import_image_advanced: fix report render

simahawk:
Left over from v14 migration.


simahawk:
Fast tracking

/ocabot merge patch

PR #161, opened 2022-01-15
[14.0][ADD] public_category relations

dnplkndll:
Odoo public product categories allow for multi site and multi category integration. provides storage support on that model.


dnplkndll:
@JordiBForgeFlow I think the idea on here was to not require the website and use the simple categories. but in our application with many sites and categories on the same products the public categories seems to make sense. So I made this as a sample of how that would look. Likely would need separated and tests added to prevent forcing the website_sale dependency.

PR #169, opened 2022-02-15
storage_file: refresh base url on startup

simahawk:
To boost URL computation for files, the base URL is stored.
Make sure it's refreshed on restart in case env settings changed
(typical case: move prod db to state/integration/dev).

### DIFF
--- a/storage_file/models/storage_backend.py
+++ b/storage_file/models/storage_backend.py
@@ -148,3 +148,8 @@ class StorageBackend(models.Model):
         else:
             parts = [backend.base_url_for_files or "", storage_file.relative_path or ""]
         return "/".join([x.rstrip("/") for x in parts if x])
+
+    def _register_hook(self):
+        super()._register_hook()
+        self.search([]).action_recompute_base_url_for_files()
+        _logger.info("storage.backend base URL for files refreshed")


storage_file/models/storage_backend.py
```diff
@@ -148,3 +148,8 @@ def _get_url_for_file(self, storage_file):
         else:
             parts = [backend.base_url_for_files or "", storage_file.relative_path or ""]
         return "/".join([x.rstrip("/") for x in parts if x])
+
+    def _register_hook(self):
```
ivantodorovich, line 152:
```suggestion
    def _register_hook(self):
        # To boost URL computation for files, the base URL is stored.
        # Make sure it's refreshed on restart in case env settings changed
        # (typical case: move prod db to state/integration/dev).
```

ivantodorovich (approved):

storage_file/models/storage_backend.py
ivantodorovich: I'm wondering then, rather than having it stored, wouldn't it make sense to `ormcache` it instead?

simahawk: no, because the storage.file url computation must be fast. It won't be as fast as having this store.
Also, invalidating the cache is hard here because we depend on server env which won't trigger any event if the value changes.

mmequignon (approved):
LGTM

storage_file/models/storage_backend.py
sebalix: Should we set a `SELECT FOR UPDATE` before updating the storage backend? We have some concurrent update issues on `storage_backend` when starting Odoo instances (too many workers starting at once and updating the same record I guess).

PR #170, opened 2022-02-21
[14.0][FIX] storage_file: lock backends when refreshing base url on startup

sebalix:
We get a lot of concurrent update errors when starting up Odoo instances lately, the workers seem to restart again and again sometimes during several minutes:
```
bad query: UPDATE "storage_backend" SET "write_uid"=1,"write_date"=(now() at time zone 'UTC') WHERE id IN (1, 2, 3, 4, 5)
ERROR: could not serialize access due to concurrent update
```

Fixing https://github.com/OCA/storage/pull/169

### DIFF
--- a/storage_file/models/storage_backend.py
+++ b/storage_file/models/storage_backend.py
@@ -151,5 +151,12 @@ class StorageBackend(models.Model):
 
     def _register_hook(self):
         super()._register_hook()
-        self.search([]).action_recompute_base_url_for_files()
+        backends = self.search([]).filtered(
+            lambda x: x._get_base_url_for_files() != x.base_url_for_files
+        )
+        if not backends:
+            return
+        sql = f"SELECT id FROM {self._table} WHERE ID IN %s FOR UPDATE"
+        self.env.cr.execute(sql, (tuple(backends.ids),), log_exceptions=False)
+        backends.action_recompute_base_url_for_files()
         _logger.info("storage.backend base URL for files refreshed")


storage_file/models/storage_backend.py
```diff
@@ -151,5 +151,8 @@ def _get_url_for_file(self, storage_file):
 
     def _register_hook(self):
         super()._register_hook()
-        self.search([]).action_recompute_base_url_for_files()
+        backends = self.search([])
```
simahawk: we could maybe filter this like:
```py
self.serch([]).filtered(lambda x: x._get_base_url_for_files() != x.base_url_for_files)
```
to avoid computing at all if not needed.

sebalix: Done

storage_file/models/storage_backend.py
```diff
@@ -151,5 +151,10 @@ def _get_url_for_file(self, storage_file):
 
     def _register_hook(self):
         super()._register_hook()
-        self.search([]).action_recompute_base_url_for_files()
+        backends = self.search([]).filtered(
+            lambda x: x._get_base_url_for_files() != x.base_url_for_files
+        )
+        sql = f"SELECT id FROM {self._table} WHERE ID IN %s FOR UPDATE"
+        self.env.cr.execute(sql, (tuple(backends.ids),), log_exceptions=False)
+        backends.action_recompute_base_url_for_files()
```
simahawk, line 161:
```suggestion
        if not backends:
            return
        sql = f"SELECT id FROM {self._table} WHERE ID IN %s FOR UPDATE"
        self.env.cr.execute(sql, (tuple(backends.ids),), log_exceptions=False)
        backends.action_recompute_base_url_for_files()
```

simahawk (changes requested):

sebalix:
@simahawk after some tests, I confirm there is no more issue when booting several Odoo instances on the same database with this PR, can you update your review?

simahawk (approved):

hparfr (approved):

PR #173, opened 2022-03-04
[14.0] storage_image_product: ease main image lookup override

sebastienbeau:
add the possibility to change the logic for getting main image, like additionnal filter based on a tag


hparfr (approved):
LGTM

simahawk (approved):
@sebastienbeau next time pls keep commit msg short and move details to its long description.
Also would be nice if the module name is mentioned.
Merci beacoup :)

https://www.odoo.com/documentation/12.0/developer/reference/guidelines.html#commit-message-structure

https://cbea.ms/git-commit/#why-not-how

PR #148, opened 2021-12-06
[MIG][15.0] storage_backend_sftp

i-vyshnevska:
depends on https://github.com/OCA/storage/pull/145

PR #178, opened 2022-04-06
[FIX] storage_backend_ftp: use full path in ``move_files()``

SilvioC2C:


storage_backend_ftp/components/ftp_adapter.py
```diff
@@ -150,8 +151,8 @@ def move_files(self, files, destination_path):
                     _logger.debug("destination %s is free", dest_file_path)
                 if result:
                     client.delete(dest_file_path)
-                # Move the file
-                client.rename(ftp_file, dest_file_path)
+                # Move the file using absolute filepaths
+                client.rename(fp(ftp_file), fp(dest_file_path))
```
simahawk, line 155: watch out: existing services could rely on the fact that the fullpath is passed already.
I would improve `_fullpath` to check if the base path is already there or not.

simahawk (changes requested):

storage_backend_ftp/components/ftp_adapter.py
SilvioC2C: @simahawk right, I'll update `_fullpath()` too to check that out

simahawk (approved):

simahawk:
we should do the same for _sftp I think

SilvioC2C:
@simahawk done here: https://github.com/OCA/storage/pull/148/commits/72ac0b61c13cc8ccf5c03a510ffddec246bec143

simahawk:
where is this commit? :thinking: :smile:

SilvioC2C:
@simahawk in the proper PR: https://github.com/OCA/storage/pull/148

simahawk:
/ocabot merge patch

Could you backport both to v14 pls?

PR #182, opened 2022-05-18
[15.0][FIX] storage_backend_sftp: Fix paramiko usage

jcoux:
Previous code used the paramiko client open function with explicit binary mode.

But, the explicit binary mode is now ignored by python.
See the `open` function description into the paramiko documentation:
https://docs.paramiko.org/en/stable/api/sftp.html

```
The Python 'b' flag is ignored, since SSH treats all files as binary.
```

But in fact, depending of the SFTP server we have an `OSError`.

Then this PR remove the useless binary mode, and avoid to have errors.

### DIFF
--- a/storage_backend_sftp/components/sftp_adapter.py
+++ b/storage_backend_sftp/components/sftp_adapter.py
@@ -76,14 +76,14 @@ class SFTPStorageBackendAdapter(Component):
                         sftp_mkdirs(client, dirname)
                     else:
                         raise  # pragma: no cover
-            remote_file = client.open(full_path, "w+b")
+            remote_file = client.open(full_path, "w")
             remote_file.write(data)
             remote_file.close()
 
     def get(self, relative_path, **kwargs):
         full_path = self._fullpath(relative_path)
         with sftp(self.collection) as client:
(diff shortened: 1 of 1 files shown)


simahawk:
@jcoux thanks! Same remark regarding the commit msg as on https://github.com/OCA/edi/pull/589 :wink:

jcoux:
> @jcoux thanks! Same remark regarding the commit msg as on [OCA/edi#589](https://github.com/OCA/edi/pull/589) wink

Done and added backport for 14.0: https://github.com/OCA/storage/pull/183

yankinmax (approved):

simahawk:
@jcoux is this specific to most recent paramiko versions? Should we keep in mind backward compat?

jcoux:
> @jcoux is this specific to most recent paramiko versions? Should we keep in mind backward compat?

I think we don't need backward compatiblity.

If you look the code of the `open` function in version  `1.16`, then since 2016:
https://github.com/paramiko/paramiko/blob/1.16/paramiko/sftp_client.py#L293
You can see the explicit binary mode was already useless.

PR #184, opened 2022-05-27
[14.0] Sync and switch to GH actions

simahawk:


hparfr (approved):

simahawk:
`storage_image_product_brand (LGPL-3) depends on product_brand (AGPL-3)`

Which one do we change? :)

simahawk:
Companion PR https://github.com/OCA/brand/pull/117

simahawk:
https://github.com/OCA/brand/pull/118

simahawk:
@hparfr can you check https://github.com/OCA/brand/pull/118 pls ?

sebastienbeau:
@simahawk you can move back storage_image_product_brand to AGPL-3.

AGPL-3 is more restrictive, but also protect us (as OpenSource contributor, by forcing other to contribute)
a good reading (GNU do not recommend to use LGPL anymore https://www.gnu.org/licenses/why-not-lgpl.en.html#top)

simahawk:
Done here https://github.com/OCA/storage/pull/207

PR #206, opened 2022-07-29
[14.0][IMP] storage_backend_ftp: Adapt move_files for ftp

rousseldenis:


rousseldenis:
As ftplib is raising a perm_error "550 Failed to change directory", catch also that Exception in order to create it.

PR #215, opened 2022-10-17
[FIX] storage_backend_sftp: Use standard put command

rousseldenis:
As open() command in write mode needs sftp user to have write access to the real file path in some operating systems, use the more common put() command to do it.


simahawk:
Is this something to fwd port?
Asking for https://github.com/OCA/storage/pull/241

Issue #67, opened 2020-10-22
[13.0] Dependency boto3 issue storage_backend_s3 


HviorForgeFlow:
Installing boto3 pip dependency with no version assigned is using version boto3==1.16.2
This version required urllib3>=1.25.4 
requests pip package has a dependency that does not allow to use urllib3 version higher than 1.25
```
Requirement already satisfied: certifi>=2017.4.17 in /usr/local/lib/python3.6/site-packages (from requests->odoo==13.0) (2020.6.20)
Collecting urllib3<1.25,>=1.21.1
  Downloading urllib3-1.24.3-py2.py3-none-any.whl (118 kB)
``` 
Once you try to update storage_backend_s3 module you will get:
```
NameError: name 'boto3' is not defined

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/odoo/custom/src/odoo/odoo/modules/registry.py", line 86, in new
    odoo.modules.load_modules(registry._db, force_demo, status, update_module)
  File "/opt/odoo/custom/src/odoo/odoo/modules/loading.py", line 419, in load_modules
    force, status, report, loaded_modules, update_module, models_to_check)
  File "/opt/odoo/custom/src/odoo/odoo/modules/loading.py", line 315, in load_marked_modules
    perform_checks=perform_checks, models_to_check=models_to_check
  File "/opt/odoo/custom/src/odoo/odoo/modules/loading.py", line 225, in load_module_graph
    load_data(cr, idref, mode, kind='data', package=package, report=report)
  File "/opt/odoo/custom/src/odoo/odoo/modules/loading.py", line 68, in load_data
    tools.convert_file(cr, package.name, filename, idref, mode, noupdate, kind, report)
  File "/opt/odoo/custom/src/odoo/odoo/tools/convert.py", line 736, in convert_file
    convert_xml_import(cr, module, fp, idref, mode, noupdate, report)
  File "/opt/odoo/custom/src/odoo/odoo/tools/convert.py", line 803, in convert_xml_import
    obj.parse(doc.getroot())
  File "/opt/odoo/custom/src/odoo/odoo/tools/convert.py", line 721, in parse
    exc_info[2]
  File "/opt/odoo/custom/src/odoo/odoo/tools/pycompat.py", line 13, in reraise
    raise value.with_traceback(tb)
  File "/opt/odoo/custom/src/odoo/odoo/tools/convert.py", line 712, in parse
    self._tag_root(de)
[...]

PR #218, opened 2022-10-19
[15.0][IMP] boto3 version bump

MiquelRForgeFlow:
After checking in detail, the problem (in https://github.com/OCA/storage/issues/67) comes with `boto3 >= 1.16.*`. Thus, let's put the latest `1.15.*` version.


AaronHForgeFlow (approved):

LoisRForgeFlow (approved):
I don't see inconvenience.

PR #235, opened 2023-02-02
[MIG] storage_backend_s3: Migration to 16.0

vince-dynapps:


storage_backend_s3/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+    "name": "Storage Backend S3",
+    "summary": "Implement amazon S3 Storage",
+    "version": "16.0.1.0.0",
+    "category": "Storage",
+    "website": "https://github.com/OCA/storage",
+    "author": " Akretion, Odoo Community Association (OCA)",
+    "license": "LGPL-3",
+    "installable": True,
+    "external_dependencies": {"python": ["boto3"]},
+    "depends": ["storage_backend"],
```
abdounasser202, line 15: Something confusing for me

In issue https://github.com/OCA/storage/issues/251, @lmignon wrote that storage_backend addon will be renamed in fs_storage, and we can see in 16.0 branch we have fs_storage

So, why in this PR **storage_backend_s3** still depends on **storage_backend** instead of **fs_backend** ?

lmignon: @abdounasser202 Since the development of the new approach based of fsspecs takes a lot of time, I decided to keep storage_* addons in 16 next to the new fs_* addons even if they cover the same functionalities... It might also be a good idea to facilitate migration from one to another and not block projects that need these features now without being able to wait for the fs_* modules to be finalised.

abdounasser202: Thanks, @lmignon 
then as I'm migrating storage_backend_ftp, it will then be better to have one depending on storage_backend and other one depending on fs_storage to keep both ?

lmignon: > Thanks, @lmignon then as I'm migrating storage_backend_ftp, it will then be better to have one depending on storage_backend and other one depending on fs_storage to keep both ?

@abdounasser202 

`fs_storage` supports natively ftp/ sftp / , S3, azure, DB, ... backends since it uses the fsspec python lib as client layer to communicate with external storage. We don't need additionnal addon by storage we want to support.

lmignon:
@vince-dynapps What's the status of this PR?

Issue #251, opened 2023-03-27
[16.0] Storage Addons Refactoring RFC


lmignon:
This RFC proposes a refactoring of the storage backend addons to make them based on Odoo standard model (`ir.attachment`) instead of custom model.

In the following, the `storage_backend` addon will be renamed in `fs_storage`  and `storage.backend` will be renamed in `fs.storage`.

## Motivation

In the `storage_file` mainly provides a custom model `storage.file` which is used to store the file content and metadata into a external file system.

Both share a lot of same fields. The not shared fields are mainly related to specific functionality provided by either `storage.file` or `ir.attachment`.

Here it's a table with the fields of the `storage.file` model and the corresponding field in the `ir.attachment` model:

| storage.file            | ir.attachment                              |
|-------------------------|--------------------------------------------|
| name                    | name                                       |
| url (computed)          | url                                        |
| url_path (computed)     |                                            |
| internal_url (computed) |                                            |
| slug (computed)         |                                            |
| relative_path           |                                            |
| file_size               | file_size                                  |
| human_file_size         | file_size (computed)                       |
| checksum                | checksum                                   |
| filename (computed)     |                                            |
| extension (computed)    |                                            |
| mimetype                | mimetype                                   |
| data                    | db_datas, raw (computed), datas (computed) |
| to_delete               |                                            |
| active                  |                                            |
| company_id              | company_id                                 |
| file_type               |                                            |
 |                         | description                                |
 |                         | res_name                                   |
|                         | res_model                                  |
|                         | res_field                                  |
|                         | res_id                                     |
|                         | type                                       |
|                         | public                                     |
|                         | access_token                               |
|                         | store_fname                                |
|                         | index_content                              |

The functional differences are:

- `storage.file` allows to access to the file content via 2 URLs stored into   the database: `internal_url` and `url`. The `url` is the public URL to access  to the file content. The `internal_url` is the URL to access to the file  content from the same  0doo's server. The `url` refers to the file into the external  file system.
- `storage.file` is used to transparently store the file content into an  external file system. When you manually create a record in the `storage.file` you can choose the external file system where the file content will be stored.
- `storage.file` is not transactional. (If a file is stored withing a transaction and the transaction is rolled back, the file will not be deleted from the external file system).
- `ir.attachment` allows to store the file content into the Odoo's filestore.
- `ir.attachment` is transactional. (If a file is stored withing a transaction and the transaction is rolled back, the file will be deleted from the filestore by a GC mechanism).
[...]

PR #252, opened 2023-03-28
[16.0][ADD] fs_storage: Storage Addons Refactoring RFC

lmignon:
A PR to track changes to the initial proposal of #251

In this PR you can find a document describing the proposal and the implementation of the first part trough the addon `fs_storage`.

The pr https://github.com/OCA/storage/pull/260 implements the new addon `fs_attachment`
The pr #261 implements the new addon `fs_file`


florian-dacosta:
@lmignon Is this ready to test/review ? (I am asking because of the draft state)

This replace storage_backend + storage_backend submodules, right ? But storage_backend is already merged.
The idea is to keep these conccurent modules for v16 and drop the storage_backend in v17 ?

lmignon:
@florian-dacosta This addon is still in draft since I'm waiting the feedback from C2C (https://github.com/camptocamp/odoo-cloud-platform/issues/418) to relicence to LGPL.
* This addon will replace the `storage_backend` addon 
* #260 fs_attachment will replace `base_attachment_object_storage` and the companion addons from https://github.com/camptocamp/odoo-cloud-platform
* #261 fs_file will replace the `storage_file` addon

I'll try to finalize #262 in the coming days but I'm busy with others priorities even if #262 is a priority. I've a more complete implementation for the  `fs_file` addon on my machine with the new widget etc...

florian-dacosta:
@lmignon Thanks for the feedback.
But this PR (`fs_storage`) does not really depend on the C2C module, so this one should be ready, no ?

I'll try to use this fs_storage in a project using  presently storage_backend (but not storage_file) and do a review here then.

lmignon:
> @lmignon Thanks for the feedback. But this PR (`fs_storage`) does not really depend on the C2C module, so this one should be ready, no ?


@florian-dacosta You're right, I'll change the status. Before I'll update the  `development_status` to alpha while others companion addons are not finalized (even if I'm pretty sure that the API should not change)

fs_storage/models/fs_storage.py
```diff
@@ -0,0 +1,426 @@
...
+        if self.protocol == "odoofs":
+            options["odoo_storage_path"] = self._odoo_storage_path
+        options = self._recursive_add_odoo_storage_path(options)
+        fs = fsspec.filesystem(self.protocol, **options)
+        directory_path = self.directory_path
+        if directory_path:
+            fs = fsspec.filesystem("rooted_dir", path=directory_path, fs=fs)
+        return fs
+
+    @deprecated("Please use _get_filesystem() instead and the fsspec API directly.")
```
florian-dacosta, line 326: It is not clear for me why this method and the other after are deprecated.
These methods are helpers that could be used in several submodules.
Of course each submodule could re-implement its own method taht will 
- decode data
- create subdirectories if missing
- create the file
But why not keeping it here to avoid code duplication in those submodules ?
I guess I am missing something... @lmignon

lmignon: @florian-dacosta The API provided by the `fsspec`lib is more complete. I've the feeling that in a lot of cases, you could find into this lib, methods more suitable to your use cases.  I would like to avoid to have to maintain yet an other one api to manager a file system storage than the one provided by the lib.

fs_storage/models/fs_storage.py
```diff
@@ -0,0 +1,426 @@
...
+        super().__init__(env, ids=ids, prefetch_ids=prefetch_ids)
+        self.__fs = None
+        self.__odoo_storage_path = None
+
+    name = fields.Char(required=True)
+    code = fields.Char(
+        required=True,
+        help="Technical code used to identify the storage backend into the code."
+        "This code must be unique. This code is used for example to define the "
+        "storage backend to store the attachments vi the configuration parameter "
```
florian-dacosta: vi -> via

fs_storage/__manifest__.py
```diff
@@ -6,16 +6,17 @@
...
-    "category": "Storage",
+    "category": "FS Storage",
     "website": "https://github.com/OCA/storage",
-    "author": " Akretion, Odoo Community Association (OCA)",
+    "author": " Akretion, ACSONE SA/NV, Odoo Community Association (OCA)",
     "license": "LGPL-3",
-    "development_status": "Production/Stable",
+    "development_status": "Beta",
     "installable": True,
-    "depends": ["base", "component", "server_environment"],
```
florian-dacosta: storage_backend did depend on server_environment, probably because you usually want to have this kind of configuration depending on the environment and also because you usually need to store some secrets, which is better not to leave in clear in the database.
Is this removal intentional, because we'd prefer to have a glue module for this ?

fs_storage/models/fs_storage.py
```diff
@@ -0,0 +1,426 @@
...
+            return func1(*args, **kwargs)
+
+        return new_func1
+
+    return decorator
+
+
+class FSStorage(models.Model):
+    _name = "fs.storage"
+    _backend_name = "fs_storage"
```
florian-dacosta: I did not see it used anymore, is it still usefull ?

fs_storage/models/fs_storage.py
```diff
@@ -0,0 +1,426 @@
...
+        compute="_compute_json_options",
+        inverse="_inverse_json_options",
+    )
+    directory_path = fields.Char(
+        help="Relative path to the directory to store the file"
+    )
+
+    # the next fields are used to display documentation to help the user
+    # to configure the backend
+    options_protocol = fields.Selection(
```
florian-dacosta, line 125: I don't get why this field is usefull as we could use `protocol` to display the right `options_properties`.
Maybe I missed something, but it would avoid some code duplication between `_get_options_protocol` and `_get_protocols` and also make sure you display the description of the chosen protocol.
It is a bit strange to have to choose a protocol and then to have to choose again to see the right description in the available option tab.

florian-dacosta (approved):
LGTM
I did some tests with Odoo filestore, gtp, sftp and it went quite well.
Some minor remarks but nothing important, so I approve already.
I have tested with a real implementation (module `attachment_synchronize`)

Thanks for the work !

fs_storage/__manifest__.py
lmignon: @florian-dacosta it's an open question... I tend to prefer to have glue addon than having too much dependencies.

fs_storage/models/fs_storage.py
lmignon: Good catch

fs_storage/models/fs_storage.py
lmignon: @florian-dacosta  With the lib you can nest the protocol. Therefore, the same field can be used to get access to the protocol documentation for the main protocol and when configuring nested protocols into the `options` field.

fs_storage/__manifest__.py
florian-dacosta: Well, before, we'd have different fields for each protocol (sftp_login, ftp_login, etc). Including server-env was avoiding a lot of small server-env glue module.
Now, we would just need one, since every thing is in the json field and we don't have extra fields per protocol.
So, no problem for me to have a glue module.

PR #256, opened 2023-04-04
[13.0][FIX] Prevent duplicated upload when uploading images using drag and drop.

benwillig:
When using the drag and drop to upload images, the _render method is recalled after files have been uploaded. This causes duplicated upload afterward.

To reproduce this issue:
- Upload a file using drag and drop
- 1 file uploaded
- Upload a new file
- the new file is uploaded twice which creates two storage.image
-  Upload a new file
- the new file is uploaded 3 times which creates three storage.image

### DIFF
--- a/storage_image/static/src/js/image_handle.js
+++ b/storage_image/static/src/js/image_handle.js
@@ -4,9 +4,18 @@ odoo.define("storage_image.image_handle", function(require) {
     var relational_fields = require("web.relational_fields");
 
     var FieldImageHandle = relational_fields.FieldOne2Many.extend({
+        init: function() {
+            this._super.apply(this, arguments);
+            this.dnd_events_added = false;
+        },
+
         _render: function() {
             var self = this;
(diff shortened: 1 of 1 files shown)


storage_image/static/src/js/image_handle.js
```diff
@@ -22,6 +32,7 @@ odoo.define("storage_image.image_handle", function(require) {
                     e.stopPropagation();
                     self.upload_images(e.originalEvent.dataTransfer.files);
                 });
+                self.events_added = true;
```
rousseldenis: @benwillig No ?

```suggestion
                self.dnd_events_added = true;
```

jdoutreloux (approved):

storage_image/static/src/js/image_handle.js
benwillig: Yes, I wanted to see who would see this. Congrats @rousseldenis

PR #260, opened 2023-04-13
[16.0][ADD] fs_attachment: Add new addon

lmignon:
This PR is part of the RFC #252 and depends on it.


fs_attachment/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+    "license": "AGPL-3",
+    "development_status": "Beta",
+    "category": "Knowledge Management",
+    "depends": ["fs_storage"],
+    "website": "https://github.com/OCA/storage",
+    "data": [
+        "security/fs_file_gc.xml",
+        "views/fs_storage.xml",
+    ],
+    "external_dependencies": {"python": ["python_slugify"]},
```
yvaucher, line 19: I don't see where slugify is used.

To remove?

requirements.txt
```diff
@@ -0,0 +1,2 @@
+# generated from manifests external_dependencies
+python_slugify
```
yvaucher, line 2: To remove if slugify is not needed.

yvaucher (approved):
Only a question about the dependency. Otherwise LGTM

requirements.txt
lmignon: it's still used here https://github.com/OCA/storage/pull/260/files#diff-43d33935418dee68316bd3b999b388a0ebc5f82bd5314bd2e6ca63352eff9965R15

lmignon:
superseded by #269

PR #262, opened 2023-05-15
[14.0][FIX] storage_backend_sftp: fix move_files

florentx:
This is an oversight. This method should take into account configuration `self.collection.directory_path` when moving files.


sebalix (approved):

simahawk (approved):
Seems legit

PR #263, opened 2023-06-02
[FIX] storage_file: fix copy

etobella:
Copy will only work when using name_id strategy, when using checksum, it will fail, as the relative_path must be unique by storage

I will port it to 15 and 16 once it has been merged

### DIFF
--- a/storage_file/models/storage_file.py
+++ b/storage_file/models/storage_file.py
@@ -44,7 +44,9 @@ class StorageFile(models.Model):
     slug = fields.Char(
         compute="_compute_slug", help="Slug-ified name with ID for URL", store=True
     )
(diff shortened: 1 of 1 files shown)


LoisRForgeFlow (approved):
:+1:

storage_file/models/storage_file.py
```diff
@@ -58,7 +60,11 @@ class StorageFile(models.Model):
     )
     mimetype = fields.Char("Mime Type", compute="_compute_extract_filename", store=True)
     data = fields.Binary(
-        help="Datas", inverse="_inverse_data", compute="_compute_data", store=False
+        help="Datas",
```
simahawk: as you are there.... :)
```suggestion
        help="Data",
```

simahawk (approved):

PR #269, opened 2023-08-01
[16.0] [IMP] fs_attachment: store attachments linked to different model/fields to different FS storages

marielejeune:
Supersedes https://github.com/OCA/storage/pull/260 (This PR is part of the RFC https://github.com/OCA/storage/pull/252 and depends on it.)

One may want to store attachments linked to different Odoo models/fields to different FS storages.
This is now possible: on the `fs.storage` one just has to configure for which models / fields it will be the default storage.
It is also possible to configure it in a server environment file,  by giving a comma-separated list of models/fields XML ids.


xavierpiernas:
I've encountered an issue with fs_attachment. After installation, I'm unable to create new attachments as it triggers the "_compute_internal_url" function, which crashes at the following line:

`
filename, extension = os.path.splitext(rec.name)
`

This occurs because the attachment initially lacks a name.

PR #271, opened 2023-08-14
[16.0][MIG] storage_image_product

sbejaoui:


lmignon:
@sbejaoui We are rewriting these addons to be based on fs_file https://github.com/OCA/storage/pull/261

PR #283, opened 2023-10-04
[16.0][FIX] fs_file: guess mimetype based on name

glitchov:
Fixes a confusion of method to guess the mimetype of a field based on its name. 
Use of mimetypes.guess_type instead of odoo.tools.mimetypes.guess_mimetype (guess on content).


lmignon:
superseded by #284 to add tests and additional logic

PR #287, opened 2023-10-13
[16.0][MIG] storage_backend_s3

ChrisOForgeFlow:
@ForgeFlow


sbidoul:
If possible for you I'd recommend migrating to fs_storage, fs_image, etc. The migration path should be quite simple.

PR #288, opened 2023-10-16
[FIX] fs_file: Browse attachment with sudo to avoid read access errors

rousseldenis:
@lmignon


lmignon:
@rousseldenis Can you add a file named `288.bugfix` into the `fs_file/readme/newsfragment` directory and describe the change and why this change.

lmignon (approved):

PR #289, opened 2023-10-16
[FIX] Missing sudo for technical models. Basic users not able to add attachments

adrienpeiffer:
Basics users without settings access are not able to upload attachment

### DIFF
--- a/fs_attachment/tests/test_fs_attachment.py
+++ b/fs_attachment/tests/test_fs_attachment.py
@@ -340,3 +340,40 @@ class TestFSAttachment(TestFSAttachmentCommon):
         self.assertEqual(attachment.checksum, attachment.store_fname.split("/")[-1])
         self.assertEqual(attachment.checksum, attachment.fs_url.split("/")[-1])
         self.assertEqual(attachment.mimetype, "text/plain")
+
+    def test_create_attachments_basic_user(self):
+        demo_user = self.env.ref("base.user_demo")
+        demo_partner = self.env.ref("base.partner_demo")
+        self.temp_backend.use_as_default_for_attachments = True
+        # Ensure basic access
+        group_user = self.env.ref("base.group_user")
+        group_partner_manager = self.env.ref("base.group_partner_manager")
+        demo_user.write(
+            {"groups_id": [(6, 0, [group_user.id, group_partner_manager.id])]}
+        )
+        # Create basic attachment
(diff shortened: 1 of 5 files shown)


fs_storage/models/fs_storage.py
```diff
@@ -163,7 +163,7 @@ def write(self, vals):
     @tools.ormcache()
     def get_id_by_code_map(self):
         """Return a dictionary with the code as key and the id as value."""
-        return {rec.code: rec.id for rec in self.search([])}
+        return {rec.code: rec.id for rec in self.sudo().search([])}
```
lmignon, line 166: @adrienpeiffer It's strange to have a `sudo` here but not in others methods.  We should add tests to be sure that we cover all the cases.

lmignon (changes requested):
@adrienpeiffer Thank you for the fix. It would be nice to have unittest for such cases. Can you also add a file named `289.bugfix` into the `fs_attachment/readme/newsfragment` and `fs_storage/readme/newsfragment` directories and describe the change and why this change. These files are used to maintain a human readable history of changes between versions.

rousseldenis (approved):
Code and functional review

fs_attachment/tests/test_fs_attachment.py
```diff
@@ -340,3 +340,40 @@ def test_storage_use_filename_obfuscation(self):
         self.assertEqual(attachment.checksum, attachment.store_fname.split("/")[-1])
         self.assertEqual(attachment.checksum, attachment.fs_url.split("/")[-1])
         self.assertEqual(attachment.mimetype, "text/plain")
+
+    def test_create_attachments_basic_user(self):
```
lmignon, line 344: :heart:

lmignon (commented):
Thank you @adrienpeiffer for the fix and the test.

Can you squash the commit into one and rewrite the commit message. Since a lot of addons are managed into the same repo, it's a good habit to start the commit with the name of the impacted modules. Something like:

```
[FIX] fs_storage, fs_attachement: Missing sudo for technical models

Before this change, users with basic access rights was no more able to create an attachment. The problem was caused by the need to access system information to determine in which file system the attachment should be created, when this sensitive information is not accessible to everyone. The change consists of reading this information in sudo to bypass this security restriction in this specific context.

```

lmignon (approved):

PR #301, opened 2023-11-17
[14.0] storage_backend: fix bug when server_env

hparfr:
has_validation is in the view; so it's computed
at creation, before the backend_type is set

The default value is not set in some circonstances


simahawk (approved):
LG

PR #303, opened 2023-11-24
fs_storage: add possibility to define options values from env.

adrienpeiffer:


fs_storage/models/fs_storage.py
```diff
@@ -302,6 +307,28 @@ def _recursive_add_odoo_storage_path(self, options: dict) -> dict:
...
+        values = {}
+        for key, value in self.json_options.items():
+            if value.startswith("$"):
+                env_variable_name = value[1:]
+                env_variable_value = os.getenv(env_variable_name)
+                if env_variable_value is not None:
+                    values[key] = env_variable_value
+                else:
+                    _logger.warning(
+                        f"Environment variable {env_variable_name} is not set."
```
sbidoul: Give more context (e.g. display_name of the fs.storage record).

sbidoul: Also, keep the env var value, instead of removing the key.

fs_storage/models/fs_storage.py
```diff
@@ -302,6 +307,28 @@ def _recursive_add_odoo_storage_path(self, options: dict) -> dict:
             self._recursive_add_odoo_storage_path(target_options)
         return options
 
+    def _eval_options_from_env(self):
+        values = {}
+        for key, value in self.json_options.items():
+            if value.startswith("$"):
```
sbidoul: Value may not be a string. If this is JSON it could be a list or a dict, or a number.
Recursive dict should be handled, I think. In the help text option there is such an example.

fs_storage/models/fs_storage.py
```diff
@@ -116,6 +116,11 @@ def __init__(self, env, ids=(), prefetch_ids=()):
         compute="_compute_json_options",
         inverse="_inverse_json_options",
     )
+
+    eval_options_from_env = fields.Boolean(
+        string="Evaluate options values starting with $ from environment variables"
```
sbidoul: That is a very long label :)
```suggestion
        string="Resolve env vars",
        help="Resolve options values starting with $ from environment variables"
```

sbidoul (approved):

sbidoul:
Looks good (code review). This should simplify configuration by letting us configure everything in a record with embedded environment variables such as, say,  `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`.

gbrito (approved):

fs_storage/models/fs_storage.py
```diff
@@ -116,6 +116,12 @@ def __init__(self, env, ids=(), prefetch_ids=()):
         compute="_compute_json_options",
         inverse="_inverse_json_options",
     )
+
+    eval_options_from_env = fields.Boolean(
```
simahawk, line 120: shouldn't be this flag controlled via server env settings?

fs_storage/models/fs_storage.py
```diff
@@ -302,6 +308,32 @@ def _recursive_add_odoo_storage_path(self, options: dict) -> dict:
...
+            elif isinstance(value, str) and value.startswith("$"):
+                env_variable_name = value[1:]
+                env_variable_value = os.getenv(env_variable_name)
+                if env_variable_value is not None:
+                    values[key] = env_variable_value
+                else:
+                    values[key] = value
+                    _logger.warning(
+                        f"Environment variable {env_variable_name} is not set for "
+                        f"fs_storage {self.display_name}."
```
simahawk, line 325:
```suggestion
                    _logger.warning(
                        "Environment variable %s is not set for "
                        "fs_storage %s", env_variable_name, self.display_name
```

fs_storage/models/fs_storage.py
```diff
@@ -116,6 +116,12 @@ def __init__(self, env, ids=(), prefetch_ids=()):
         compute="_compute_json_options",
         inverse="_inverse_json_options",
     )
+
+    eval_options_from_env = fields.Boolean(
+        string="Resolve env vars",
+        help="Resolve options values starting with $ from environment variables",
```
simahawk, line 122: I would provide an example here

simahawk (commented):
LG overall. I'll test later

fs_storage/models/fs_storage.py
sbidoul: The idea is to set this flag in XML records as well as `json_options` and then configure via environment variables.
Do you have other use cases in mind? Our main one is compatibility of configuration with `attachment_s3`.

simahawk: I'm just thinking that if you want the env conf to take full precedence, you should be able to set it via env key too otherwise the env settings are useless if you don't have the flag set.

However, I understand that you might want to leave the control from the UI.

lmignon (commented):
Thank you @adrienpeiffer 
Can you complete the documentation in https://github.com/OCA/storage/blob/16.0/fs_storage/readme/USAGE.rst to describe this new option for configuring a storage backend?
Can you also add a file `303.feature` into the  `fs_storage/readme/newsfragment` directory to describe this new feature. This file will be use at merge to fill the history section into the readme file and help users to know what changes have been made between versions.

adrienpeiffer:
@lmignon Could be merged for Xmas ?

sbidoul:
@adrienpeiffer can you handle the two comments from Simone?

sbidoul:
I have cancelled the merge.

adrienpeiffer:
@sbidoul Seems not cancelled. I'll handle that in an other PR.

PR #304, opened 2023-11-25
[16.0][FIX] fs_attachment: Fix attachment store after upgrade

lmignon:
Fix attachments stored in Odoo after an addon upgrade.

Prerequisites:

* The fs_attachment addon is installed, and specific filesystem storage is
  configured.
* All the attachments are already stored in the appropriate filesystem storage.
  (We assume that if you've installed the addon on an existing database,
  you'vealready migrated the existing attachments appropriately.

Context:

* You run an upgrade command on your Odoo server.

Problem:

Some attachments could be created during the upgrade process before the load of
the fs_attachment module or the configuration of the filesystem storage backends.
This is especially true for the attachments created by addons to store the app
icons or even attachments from XML data files. As a result, you end up with
attachments that could not be found by some Odoo instances if, for example, you have
multiple Odoo instances running on different servers since these attachments are stored
in the odoo filestore of the server that created them.

Solution:

A solution to this problem is to hook the end of the upgrade process through the
implementation of the *_register_hook* method on our *ir.attachment* model. Since
this method is called after the completion of the upgrade process and the full
initialization of the odoo registry, we can safely expect that all the information
we need to know where to store the attachments are available.
From there, we can simply search for all the attachments that are not linked to
any specific storage and store them in the appropriate storage.


lmignon:
cc @sbidoul @adrienpeiffer

fs_attachment/models/ir_attachment.py
```diff
@@ -96,6 +79,46 @@ class IrAttachment(models.Model):
...
+        calframe = inspect.getouterframes(curframe, 2)
+        # the caller of _register_hook is 'load_modules' in
+        # odoo/modules/loading.py
+        load_modules_frame = calframe[1][0]
+        # 'update_module' is an argument that 'load_modules' receives with a
+        # True-ish value meaning that an install or upgrade of addon has been
+        # done during the initialization. We need to move the attachments that
+        # could have been created or updated in other addons before this addon
+        # was loaded
+        update_module = load_modules_frame.f_locals.get("update_module")
```
sbidoul, line 111: Is it not too much magic? I don't know, I think I prefer to do these kind of operations that can potentially lose data explicitly.

len-foss: Couldn't this be simplified by simply running it every time?
If there's nothing to do the cost shouldn't be too expensive.

The manual solution has the issue to be an additional hurdle for deployment that requires to be well documented, or the user might face some troubles that might be bad for adoption.

CasVissers-360ERP:
@lmignon @sbidoul are you aware that the base vacuuming of Odoo will never run, also for the ignored files?

https://github.com/odoo/odoo/blob/e721ed146d6173cc04efe6103aa6e682fb62917f/odoo/addons/base/models/ir_attachment.py#L165

CasVissers-360ERP:
In addition I think there is an issue with force_storage #310

PR #305, opened 2023-11-29
[FIX] fs_image: remove rawCacheKey invalidation

sbejaoui:
Fix view crash when uploading an image

The rawCacheKey is appropriately managed by the base class and reflects the
record's last update datetime (write_date).
Since it lacks a setter, attempting to invalidate its value results in a view crash.
Nevertheless, the value will automatically be updated upon saving the record.


lmignon:
@sbejaoui  can you add a fragment to include into the addon history... as it's done here https://github.com/OCA/storage/pull/304/files#diff-0c0256bf7a5f02140833fcc325442a42553aac6aaf8b013258f347d6ce2d7e8c

fs_image/static/src/views/fields/fsimage_field.esm.js
```diff
@@ -50,9 +50,6 @@ export class FSImageField extends ImageField {
                     fileTypeMagicWordMap[this.props.value.content[0]] || "png";
                 return `data:image/${magic};base64,${this.props.value.content}`;
             }
-            if (!this.rawCacheKey) {
```
lmignon, line 53: @sbejaoui In both cases we redefine a method and don't call super. We should therefore process the 'rawCacheKey'

lmignon: OK my Odoo was not up to date. Indeed, the 'rawCacheKey' is now a property

sbejaoui:
> @sbejaoui can you add a fragment to include into the addon history... as it's done here https://github.com/OCA/storage/pull/304/files#diff-0c0256bf7a5f02140833fcc325442a42553aac6aaf8b013258f347d6ce2d7e8c

done

fs_image/readme/newsfragments/305.bugfix
```diff
@@ -0,0 +1,6 @@
+Remove rawCacheKey invalidation.
```
lmignon: @sbejaoui Thank you for this changelog.  A little question. As a non developer or without having done this change, do you think this change can help to know if you need to update your addon? IMO, the title should explain for the end user what's this PR fixes.

lmignon (changes requested):

lmignon (approved):

PR #308, opened 2023-12-03
[FIX] fs_attachment: add a few missing sudo to fs_attachment

sbidoul:
fs.storage is protected as it contains credentials. It must be accessed sudo when needed for regular users manipulating ir.attachments.

I also add some sudo in compute functions. These are currently executed sudo anyway but this may change in the future.

### DIFF
--- a/fs_attachment/models/ir_attachment.py
+++ b/fs_attachment/models/ir_attachment.py
@@ -135,7 +135,7 @@ class IrAttachment(models.Model):
         for rec in self:
             if rec.store_fname:
                 code = rec.store_fname.partition("://")[0]
-                fs_storage = self.env["fs.storage"].get_by_code(code)
+                fs_storage = self.env["fs.storage"].sudo().get_by_code(code)
                 if fs_storage != rec.fs_storage_id:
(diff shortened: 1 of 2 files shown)


fs_attachment/models/ir_attachment.py
```diff
@@ -135,7 +135,7 @@ def _compute_fs_storage_id(self):
         for rec in self:
             if rec.store_fname:
                 code = rec.store_fname.partition("://")[0]
-                fs_storage = self.env["fs.storage"].get_by_code(code)
+                fs_storage = self.env["fs.storage"].sudo().get_by_code(code)
                 if fs_storage != rec.fs_storage_id:
                     rec.fs_storage_id = fs_storage
```
sbidoul, line 140: Here we assign a sudoed record to a non-sudoed attachment. Could this be a problem? IMO no

sbidoul: Note this problem was already present before because compute_sudo is the default, so just noting this "en passant".

lmignon (approved):

PR #309, opened 2023-12-06
[ADD][16.0] fs_storage_backup

tarteo:


lmignon (approved):
Why not even If I prefer to use dedicated software to manage my backups...

LGTM (Code review only)

fs_storage_backup/models/fs_storage.py
```diff
@@ -0,0 +1,116 @@
...
+            self.fs.makedirs(self.backup_dir, exist_ok=True)
+            if self.fs.exists(backup_path):
+                raise Exception("File already exists (%s)." % backup_path)
+            backup_file = self.fs.open(backup_path, "w")
+            list_db = tools.config["list_db"]
+            if not list_db:
+                tools.config["list_db"] = True
+            db.dump_db(
+                self.env.cr.dbname,
+                backup_file.buffer,
```
lmignon, line 75: What about memory consumption. Is the buffer a temporary file?

tarteo: Here buffer is a file-like for s3 default blocksize is 5242880

tarteo:
Can we reopen this?

tarteo:
@OCA/storage-maintainers Can we make this no-stale, please?

3coma3:
@OCA/storage-maintainers Hi, just adding a bit of rationale. I agree that dedicated systems are the best for DR but this is a quick and easy resource for SaaS/resellers wanting to give customers a degree of control over their data, not really meant to replace those tools.

3coma3:
Many thanks to all!

PR #312, opened 2023-12-13
[16.0] fs_attachment: fix meaningful name rewriting

len-foss:
This fix the name rewriting in case the storage is "directory name optimized".

It adds a test to ensure that the name is consistently used between write and read.

Closes https://github.com/OCA/storage/pull/285


lmignon (commented):
Thank you @len-foss LGTM (Code review) 
I wonder if this behaviour also works correctly when using the 'file like' api to modify the contents of the attachment....

```
with attachment.open("wb") as f:
    f.write(...)
```

Nevertheless, can you add a file named `312.bugfix` into the `fs_attachment/readme/newsfragment` directory to describe the change in an 'end user 'oriented way.  This description will be add to the history section of the module description at merge

len-foss:
I'm very confused, as the newly added test breaks on the `write`, and only so if other tests from the same file (`test_fs_attachment`) are run. The way it breaks is also very strange:
```
FileNotFoundError: [Errno 2] No such file or directory: '/github/home/.local/share/Odoo/filestore/odoo/storage//tmp/tmpq6abj5qo/github/home/.local/share/Odoo/filestore/odoo/storage//te/st/test.txt'
```
Let's take a bit more time to understand where the issue is coming from, unless you already have some pointer?

lmignon:
> I'm very confused, as the newly added test breaks on the `write`, and only so if other tests from the same file (`test_fs_attachment`) are run. The way it breaks is also very strange:
> 
> ```
> FileNotFoundError: [Errno 2] No such file or directory: '/github/home/.local/share/Odoo/filestore/odoo/storage//tmp/tmpq6abj5qo/github/home/.local/share/Odoo/filestore/odoo/storage//te/st/test.txt'
> ```
> 
> Let's take a bit more time to understand where the issue is coming from, unless you already have some pointer?

The tests are failing due to the env cleanup at transaction rollback after each test. Indeed, the fs.storage backend inherit from the server.env.mixin. Therefore some fields are pseudo computed fields and need to be reset before running each test. To fix your issue, you need to update your created fs.storage record into the `setup` method....

```python

def setUp(self)::
    super().setUp()
    self.backend_optimized = write(
            {
                "protocol": "file",
                "code": "tmp_opt",
                "directory_path": temp_dir,
                "optimizes_directory_path": True,
            }
        )

```

len-foss:
Oh, thank you very much! 
I see, that also explains why I had to create the other backend, reusing the other made the tests break unexplainably in a similar fashion.

We should be good now then :-)

fs_attachment/tests/test_fs_attachment.py
```diff
@@ -30,6 +30,38 @@ def test_create_attachment_explicit_location(self):
...
+        # The storage root is NOT part of the store_fname
+        self.assertFalse("tmp/" in attachment.store_fname)
+
+        # we can read the file, so storage finds it correctly
+        with attachment.open("rb") as f:
+            self.assertEqual(f.read(), content)
+
+        new_content = b"new content"
+        with attachment.open("wb") as f:
+            f.write(new_content)
```
lmignon, line 67: IMO, after the 'write' you should also check the store_fname to be sure that the new file is till into an optimized directory path...

lmignon (approved):
A little comment otherwise LGTM (Code review).

Thank you for the contrib. Don't forget to add you name into the contributor list :smirk:

len-foss:
All right. I have added this to the test and a bit more explanations to make it easier to check the assumptions.

lmignon (approved):

PR #314, opened 2023-12-14
[16.0][IMP] fs_storage: compute options protocol + refactor test class for re-use

len-foss:
Before the change, reading the options protocol might not be in sync with the record protocol.
This is particularly the case if the record is save in the middle of editing as the value resets, which can be quite confusing.

### DIFF
--- a/fs_storage/tests/__init__.py
+++ b/fs_storage/tests/__init__.py
@@ -1,1 +1,2 @@
+from . import common
 from . import test_fs_storage
--- a/fs_storage/tests/common.py
+++ b/fs_storage/tests/common.py
@@ -1,6 +1,7 @@
 # Copyright 2023 ACSONE SA/NV (http://acsone.eu).
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
 import base64
+import os
 import shutil
 import tempfile
 from unittest import mock
@@ -21,6 +22,7 @@ class TestFSStorageCase(TransactionCase):
         cls.filename = "test_file.txt"
         cls.case_with_subdirectory = "subdirectory/here"
         cls.demo_user = cls.env.ref("base.user_demo")
+        cls.temp_dir = tempfile.mkdtemp()
 
     def setUp(self):
         super().setUp()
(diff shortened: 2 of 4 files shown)


len-foss:
@lmignon really tiny change but it makes things a bit easier, as while setting options the user might get confused between host/url/endpoint that mean the same thing but have a different key depending on the protocol.

len-foss:
I've also added a split of the test class to be able to reuse it in other modules (in particular to work on `dms_storage`).

lmignon (approved):
Thank you @len-foss LTGM
It's indeed less confusing for the end user

len-foss:
So I've pushed another commit to help test other module that might depend on `fs_storage`.
While working on https://github.com/OCA/dms/pull/275, as soon as there are more than one test that are run, the file would be created on a `tmp_dir_1` and then the very next following file read would be on `tmp_dir_2`. It can be really maddening, so I think having a test class that resist these pain points can be really helpful.

fs_storage/tests/__init__.py
```diff
@@ -1 +1,2 @@
+from . import common
```
simahawk, line 1: there's no need to import common here

fs_storage/tests/common.py
```diff
@@ -0,0 +1,46 @@
...
+
+        # pylint: disable=unused-variable
+        @self.addCleanup
+        def stop_mock():
+            mocked_backend.stop()
+            # recursively delete the tempdir
+            if os.path.exists(self.temp_dir):
+                shutil.rmtree(self.temp_dir)
+
+    def _create_file(self, backend: FSStorage, filename: str, filedata: str):
```
simahawk, line 44: I'm not sure we need this method... are we going to end up w/ all the possible methods of the backend?

len-foss: This is part of the original code that has just been moved. If you look at the individual commits, the only change is https://github.com/OCA/storage/pull/314/commits/2ad064e1cac36bd1585d2e52928597546aec40ff

It makes it easier to re-use the backend and to test with 'optimize paths', which works with subfolders.

PR #232, opened 2023-01-17
[12.0][UPD] update dotfiles using copier

AungKoKoLin1997:
In this PR, update dotfiles using copier and fix `development_status` in modules.
@qrtl

PR #315, opened 2023-12-18
[12.0][UPD] dotfiles

thomaspaulb:
Retake of https://github.com/OCA/storage/pull/232


thomaspaulb:
@sbidoul, do you have any idea why `manifestoo` complains that `server_environment` is Beta, when in fact it's [Production/Stable](https://github.com/OCA/server-env/blob/12.0/server_environment/__manifest__.py)?

sbidoul:
Ah, that is because [odoo12-addon-server-environment ](https://pypi.org/project/odoo12-addon-server-environment/12.0.2.0.5/) (the latest non-pre release and the one installed by the CI) is Beta. And the change to Production/Stable has likely been made without a version bump.

In the latest branches we don't use pre-releases anymore but for older branches it is important to bump versions.

thomaspaulb:
@sbidoul Thanks! I've merged a [bump](https://github.com/OCA/server-env/pull/174), let's see how long it takes for the PyPi version to update.

sbidoul:
@thomaspaulb looks like you used nobump instead of patch?

thomaspaulb:
Yes, because i already bumped the version manually, so as to not bump it double. Not ok?

sbidoul:
Yeah but that landed you another pre-release. You need to ocabot merge patch.

thomaspaulb:
That worked!

@OCA/storage-maintainers Can this be fast-tracked, because the CI of other PR's depends on it.

PR #317, opened 2024-01-04
[16.0][IMP] fs_storage: pattern matching should only apply to file name, not full path

tuantrantg:


fs_storage/models/fs_storage.py
```diff
@@ -408,7 +408,8 @@ def find_files(self, pattern, relative_path="", **kw) -> list[str]:
             return []
         regex = re.compile(pattern)
         for file_path in self.fs.ls(relative_path, detail=False):
-            if regex.match(file_path):
+            # fs.ls returns a relative path
```
simahawk, line 411: As this method is deprecated, does it makes sense to improve it? Unless, this is a fix...

simahawk (commented):
A remark for the commit msg
```
[IMP or FIX] fs_storage: improve/fix find_files pattern match

bla bla explanation 
```

fs_storage/models/fs_storage.py
tuantrantg: yes, it's a fix

lmignon (approved):

PR #318, opened 2024-01-11
[16.0][IMP] fs_storage: move default fs_storage from demo to data

kevinkhao:
this can be useful for some modules where you want to create some other model with a default fs_storage


bealdav (approved):
it makes sense

LGTM

lmignon (changes requested):
It's a breaking change since it'll create a new record into the databases where this addon is already installed. Moreover I don't see any benefit to have such a default record created by default since its configuration depends on your specific usage.

bealdav:
Ok thanks for your anwser.

PR #320, opened 2024-01-12
[16.0][IMP] fs_storage: invalidate filesystem cache when connection fails

### LINKED ISSUE #319: [16.0] fs_storage + EDI: connection closed after some time
## Module
fs_storage

## Describe the bug
I am using fs_storage to connect to SFTP with EDI (the PR is still not merged https://github.com/OCA/edi-framework/pull/26). The connection works fine, but after some time the EDI jobs start to fail due to the following issue:

```
File "/opt/odoo/auto/addons/fs_storage/models/fs_storage.py", line 376, in add
    self.fs.makedirs(self.fs.sep.join(path))
  File "/usr/local/lib/python3.10/site-packages/fsspec/implementations/sftp.py", line 89, in makedirs
    self.ftp.mkdir(new_path, mode)
  File "/usr/local/lib/python3.10/site-packages/paramiko/sftp_client.py", line 460, in mkdir
    self._request(CMD_MKDIR, path, attr)
  File "/usr/local/lib/python3.10/site-packages/paramiko/sftp_client.py", line 856, in _request
    num = self._async_request(type(None), t, *args)
  File "/usr/local/lib/python3.10/site-packages/paramiko/sftp_client.py", line 881, in _async_request
    self._send_packet(t, msg)
  File "/usr/local/lib/python3.10/site-packages/paramiko/sftp.py", line 209, in _send_packet
    self._write_all(out)
  File "/usr/local/lib/python3.10/site-packages/paramiko/sftp.py", line 173, in _write_all
    n = self.sock.send(out)
  File "/usr/local/lib/python3.10/site-packages/paramiko/channel.py", line 799, in send
    return self._send(s, m)
  File "/usr/local/lib/python3.10/site-packages/paramiko/channel.py", line 1196, in _send
    raise socket.error("Socket is closed")
OSError: Socket is closed
```
If I press the button 'TEST CONNECTION' in the fs storage, I also get the same error. However, if I restart Odoo, it starts working fine again.

My hint is that the problem comes from fs_storage storing the connection in a variable (https://github.com/OCA/storage/blob/3083166c5fbf4e4a888babfc916caf9f6f530034/fs_storage/models/fs_storage.py#L274). The connection closes at some point (probably due to timeout) and it does not work anymore.

@lmignon @sebastienbeau have you faced something like this with fs_storage?

JordiMForgeFlow:
Fixes https://github.com/OCA/storage/issues/319

CC @LoisRForgeFlow


fs_storage/models/fs_storage.py
```diff
@@ -130,6 +130,16 @@ def __init__(self, env, ids=(), prefetch_ids=()):
...
     )
 
+    store_connection = fields.Boolean(
+        default=True,
+        help="If marked, the connection to the filesystem is stored in memory.",
+    )
+    # https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.spec.AbstractFileSystem.clear_instance_cache  # noqa: B950
+    invalidate_cache_on_error = fields.Boolean(
+        help="If marked, the filesystem cache will be invalidated when there is an "
+        "error in the connection. Invalidating the cache will force a reconnection.",
```
lmignon: @JordiMForgeFlow  Should this be set to True by default to ensure a safe use of fspecs?

fs_storage/models/fs_storage.py
```diff
@@ -363,6 +375,12 @@ def _get_filesystem(self) -> fsspec.AbstractFileSystem:
         directory_path = self.directory_path
         if directory_path:
             fs = fsspec.filesystem("rooted_dir", path=directory_path, fs=fs)
+        if self.invalidate_cache_on_error:
+            # Check whether we need to invalidate cache or not
+            try:
+                fs.ls("", detail=False)
```
lmignon: If everything is stored at the root of the file system, this will shortly introduce a performance issue. Why not `fs.exists("\")`?

lmignon (changes requested):
@JordiMForgeFlow Thank you for this improvement. A few comments... In the same time can you add a file named `320.feature` into the fs_storage/readme/newsframent with a description of this new feature. The content will be used at merge time to fill the addon's changelog.
Could you also update the [USAGE.rst](https://github.com/OCA/storage/blob/16.0/fs_storage/readme/USAGE.rst) file to describe the new parameters?

fs_storage/models/fs_storage.py
JordiMForgeFlow: @lmignon I agree, not even sure if there will be cases where this is not wanted, but we can keep it as a setting.

fs_storage/models/fs_storage.py
JordiMForgeFlow: @lmignon true, I was using the same approach as in the method to test the configuration (not sure if maybe we could also change it there)

lmignon: > @lmignon true, I was using the same approach as in the method to test the configuration (not sure if maybe we could also change it there)

IMO we should

fs_storage/models/fs_storage.py
```diff
@@ -363,6 +376,12 @@ def _get_filesystem(self) -> fsspec.AbstractFileSystem:
         directory_path = self.directory_path
         if directory_path:
             fs = fsspec.filesystem("rooted_dir", path=directory_path, fs=fs)
+        if self.invalidate_cache_on_error:
+            # Check whether we need to invalidate cache or not
+            try:
+                fs.exists("/")
+            except Exception:
+                fs.clear_instance_cache()
```
lmignon: If it fails should we raise the error??? IMO we should.

LoisRForgeFlow (approved):
LGTM

invalidate_cache_on_error maybe is not even needed as a setting and what it adds should be just the standard behavior, but not a big deal.

fs_storage/models/fs_storage.py
JordiMForgeFlow: @lmignon changed :)

fs_storage/models/fs_storage.py
JordiMForgeFlow: @lmignon changed it

JordiMForgeFlow:
@lmignon after trying using `fs.exists` and looking into more detail, I have put back using `fs.ls`.

The `fs.exists` does not raise an error, but simply returns true or false. Also, it basically calls the method `info` which ends up calling the method `fs.ls` but with detail=True, which is worse than what we had. (code can be seen here: https://filesystem-spec.readthedocs.io/en/latest/_modules/fsspec/spec.html#AbstractFileSystem)

fs_storage/models/fs_storage.py
```diff
@@ -130,6 +130,17 @@ def __init__(self, env, ids=(), prefetch_ids=()):
         help="Relative path to the directory to store the file"
     )
 
+    store_connection = fields.Boolean(
```
simahawk: I wonder if this is needed and if yes, under which circumstances.
Also, is worth to mention in the help that the cache is only by thread.

fs_storage/models/fs_storage.py
```diff
@@ -130,6 +130,17 @@ def __init__(self, env, ids=(), prefetch_ids=()):
...
+        default=True,
+        help="If marked, the connection to the filesystem is stored in memory.",
+    )
+    # https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.spec.AbstractFileSystem.clear_instance_cache  # noqa: B950
+    invalidate_cache_on_error = fields.Boolean(
```
simahawk: Do we really need a flag for this? IMO we can assume that this is the safest thing to do.

lmignon:
> @lmignon after trying using `fs.exists` and looking into more detail, I have put back using `fs.ls`.
> 
> The `fs.exists` does not raise an error, but simply returns true or false. Also, it basically calls the method `info` which ends up calling the method `fs.ls` but with detail=True, which is worse than what we had. (code can be seen here: https://filesystem-spec.readthedocs.io/en/latest/_modules/fsspec/spec.html#AbstractFileSystem)

My concern is about the performance cost when the root directory contains hundred of thousand files... To limit the scope of the `ls` command, it could be performed on a specific path to a marker file. What do you thing about?

simahawk:
> > @lmignon after trying using `fs.exists` and looking into more detail, I have put back using `fs.ls`.
> > The `fs.exists` does not raise an error, but simply returns true or false. Also, it basically calls the method `info` which ends up calling the method `fs.ls` but with detail=True, which is worse than what we had. (code can be seen here: [filesystem-spec.readthedocs.io/en/latest/_modules/fsspec/spec.html#AbstractFileSystem](https://filesystem-spec.readthedocs.io/en/latest/_modules/fsspec/spec.html#AbstractFileSystem))
> 
> My concern is about the performance cost when the root directory contains hundred of thousand files... To limit the scope of the `ls` command, it could be performed on a specific path to a marker file. What do you thing about?

I agree. We should probably create a file on 1st check (eg: `.odoo_fs_storage_$id.marker`).

fs_storage/models/fs_storage.py
JordiMForgeFlow: @simahawk you are right, I have not checked it but probably with the invalidation of the cache of the FS it will be enough. I will remove it and test.

fs_storage/models/fs_storage.py
```diff
@@ -363,6 +366,18 @@ def _get_filesystem(self) -> fsspec.AbstractFileSystem:
         directory_path = self.directory_path
         if directory_path:
             fs = fsspec.filesystem("rooted_dir", path=directory_path, fs=fs)
+        if not tools.config["test_enable"]:
+            # Check whether we need to invalidate FS cache or not.
+            # Use a marker file to limit the scope of the LS command for performance.
+            try:
+                marker_file_name = self._get_marker_file_name()
```
simahawk: Is it really needed to make this call every time?
IMO we should:
1. move this to `_check_connection`
2. call it here only in case of connection failure to raise a specific error when possible
3. use it w/ test btn

fs_storage/models/fs_storage.py
```diff
@@ -339,6 +339,9 @@ def _get_fs_options(self):
             return options
         return self._eval_options_from_env(self.json_options)
 
+    def _get_marker_file_name(self):
+        return "odoo_fs_storage_%s.marker" % self.id
```
simahawk:
```suggestion
        return ".odoo_fs_storage_%s.marker" % self.id
```
and drop the manipulation below

simahawk (changes requested):

fs_storage/models/fs_storage.py
JordiMForgeFlow: @simahawk the problem is that I need to call it in order to know if the connection is still working

simahawk: Ok but I tend to think that if the conn works we shouldn't do this check every time.

JordiMForgeFlow: @simahawk another option would be to wrap all the methods that call FS methods in a try/except block that then invalidates the cache in case of failing.

On the other hand, I have just faced this problem with an SFTP filesystem. Maybe we could simply keep it as an FS Storage option and only use it when it is needed.

sebalix:
Hello @JordiMForgeFlow , when testing this PR, I get the following notification when using the Test Connection button:
image

Without the PR the connection is working well.

JordiMForgeFlow:
@sebalix it seems to be coming from an issue in FS. The Odoo filesystem uses the local filesystem class. There was an issue with the ls for files (https://github.com/fsspec/filesystem_spec/pull/1479). It is already merged but not released yet, I have asked them when should we expect it.

simahawk:
@sebalix can you try by installing it from https://github.com/fsspec/filesystem_spec ?

fs_storage/models/fs_storage.py
```diff
@@ -339,6 +339,17 @@ def _get_fs_options(self):
...
         return self._eval_options_from_env(self.json_options)
 
+    def _get_marker_file_name(self):
+        return ".odoo_fs_storage_%s.marker" % self.id
+
+    def _check_connection(self, fs=False):
+        if not fs:
+            fs = self.fs
+        marker_file_name = self._get_marker_file_name()
+        if not fs.ls(marker_file_name, detail=False):
```
sebalix: @JordiMForgeFlow @simahawk even by installing `fsspec@master` I get the error `FileNotFoundError: [Errno 2] No such file` on this line.

FYI it has been tested with `sftp` protocol, the issue is raised on this line exactly: https://github.com/fsspec/filesystem_spec/blob/master/fsspec/implementations/sftp.py#L127

Should we do this check inside a `try/except` instead? I even tried `self.fs.info(path)` instead of `ls` but it is raising the same `FileNotFound` exception.

sebalix: Even with `file/local` protocol (LocalFilesystem) it is failing:
`FileNotFoundError: [Errno 2] No such file or directory: '/odoo/.odoo_fs_storage_4.marker'`
A try/except is mandatory to check if a path exists.

lmignon: Why don't we use the 'exists' method?

JordiMForgeFlow: @lmignon the problem I see with 'exists' is that it returns either True or False, it covers other potential errors that could raise from the connection. I think that the 'ls' with a try/except for the 'filenotfounderror' should work. WDYT?

lmignon: :+1:

sebalix: Should we also check for `PermissionError`? E.g. with a forbidden folder:
`PermissionError: [Errno 13] Permission denied: '/root/.odoo_fs_storage_4.marker'`

JordiMForgeFlow: @sebalix how would you then handle this error? We would not be able to use the marker file, right?

sebalix: Maybe we should not handle it you are right, and let it pop-up and get an error about it.

sebalix:
Regarding connection failure, I replaced the use of the deprecated functions in `edi_storage_oca` there: https://github.com/OCA/edi-framework/pull/65#issuecomment-1923307964
Copying my comment:

> I'm suspecting (but I'm not sure) it was the use of these deprecated methods that triggers connection issues over time. These methods are decorated, and it could be that the decorator is keeping an old reference to self.__fs having a socket that gets closed after some time. That could explain why the "Test connection" button of fs_storage provides the expected result, while the EDI jobs are failing even when requeued.
This will be tested anyway.

Maybe it's not related at all.

fs_storage/models/fs_storage.py
JordiMForgeFlow: @sebalix I have added the handling for the FileNotFoundError. I have tested locally and it works fine. Could you check from your side?

sebalix: @JordiMForgeFlow now it works as expected, thanks!

fs_storage/models/fs_storage.py
```diff
@@ -339,6 +339,21 @@ def _get_fs_options(self):
             return options
         return self._eval_options_from_env(self.json_options)
 
+    def _get_marker_file_name(self):
+        return ".odoo_fs_storage_%s.marker" % self.id
+
+    def _check_connection(self, fs=False):
```
sebalix: I have some doubts: this `_check_connection` method is not called by `def fs` property, so the FS object returned by `_get_filesystem` will continue to persist in the worker memory, even if its socket gets closed over time.
Should we not call this method from `def fs` as well?

What we did to avoid such connection errors is a quick&dirty patch to always return a new FS object through `def fs` property, this adds some overhead but guarantee the connection has been refreshed.

JordiMForgeFlow: @sebalix I had the same doubt but I wasn't sure if the FS object persists for much long or not. I initially added an option to decide if you wanted to store the FS object or not.

fs_storage/models/fs_storage.py
```diff
@@ -363,6 +378,14 @@ def _get_filesystem(self) -> fsspec.AbstractFileSystem:
...
         if directory_path:
             fs = fsspec.filesystem("rooted_dir", path=directory_path, fs=fs)
+        if not tools.config["test_enable"]:
+            # Check whether we need to invalidate FS cache or not.
+            # Use a marker file to limit the scope of the LS command for performance.
+            try:
+                self._check_connection(fs)
+            except Exception as e:
+                fs.clear_instance_cache()
+                raise e
```
lmignon: IMO this code should go into the `fs` property method....

JordiMForgeFlow: @sebalix with this change now we ensure the connection is refreshed if needed

fs_storage/models/fs_storage.py
```diff
@@ -339,6 +347,21 @@ def _get_fs_options(self):
             return options
         return self._eval_options_from_env(self.json_options)
 
+    def _get_marker_file_name(self):
+        return ".odoo_fs_storage_%s.marker" % self.id
+
+    def _check_connection(self, fs=False):
+        if not fs:
```
simahawk: I would make this a model method and make `fs` a required arg.

simahawk (changes requested):

fs_storage/models/fs_storage.py
sebalix: @JordiMForgeFlow it should be fine like this from what I see in my tests :+1: seems the property is only useful within the same Odoo transaction, the `self.__fs` object doesn't persist in memory between 2 HTTP requests on Odoo (`__init__` is always called and resets `self.__fs` to None, so we always enter in the `if not self.__fs:` block in the property), so all good.

sebalix (approved):

fs_storage/models/fs_storage.py
JordiMForgeFlow: @simahawk we use 'self' to generate the marker file name, maybe better to keep it without being a model method?

simahawk: Yeah, ok, but this thing of falling back to self.__fs is not need IMO. Either we expect to pass it or we don't.
Not a blocker tho if you think it makes sense :wink:

JordiMForgeFlow: @simahawk I agree that it looks better, could you check now?

simahawk (approved):

lmignon (approved):

fs_storage/models/fs_storage.py
```diff
@@ -266,12 +266,33 @@ def _compute_options_properties(self) -> None:
...
         """Get the fsspec filesystem for this backend."""
         self.ensure_one()
         if not self.__fs:
             self.__fs = self._get_filesystem()
+        if not tools.config["test_enable"]:
+            # Check whether we need to invalidate FS cache or not.
+            # Use a marker file to limit the scope of the LS command for performance.
+            try:
+                self._check_connection(self.__fs)
+            except Exception as e:
```
sebalix, line 293: After some tests, we are still facing connections issues and the current implementation doesn't cover it: ```OSError: Socket is closed```
This errors arrives only after some time (several hours, workers have a high lifespan here and FS object is never recreated meanwhile).

JordiMForgeFlow: @sebalix maybe then we should not store the FS object at all in the slot, and always call to _get_filesystem

lmignon: The module can also be used in contexts of heavy traffic between odoo and external filesystems. This is particularly the case when it is used to manage Odoo's default attachment storage. In such circumstances, the cost of the reconnection required each time the fs proxy would be recreated is not negligible. Perhaps it would be a good idea to set up a 'keep-alive' mechanism via a cron running every 'X' minutes...

JordiMForgeFlow: @lmignon we could do that or putting it as a configuration in the FS storage to decide if you want to check the connection everytime or not. The cron approach however seems to be the solution covering both cases, and one can decide to run it more/less frequently. What do you think @sebalix ?

sebalix: Not sure to understand the cron approach regarding this in-memory object?
But yes, the only fix that is working for us right now is to return a new FS object every time.

JordiMForgeFlow: I understand it would be to have a cron that invalidates the __fs slot after some time

sebalix: The cron process could updates memory objects of other Odoo process (workers)? Not sure it's possible :thinking:

sebalix: The `raise e` is an issue here (testing the code during a workshop), it is blocking somehow the queue-job.
Clearing the fsspec cache should be enough here, so a new connection will be established in case of error.

sebalix: Ok, after one day of testing, we encountered no connection issues with the help of the `skip_instance_cache` parameter of `fsspec` (set in the options). This will add some overhead but at least we are safe... dealing with this cache in Odoo seems error prone (connection stability, configuration of the SFTP server...).

I don't know how huge is this overhead, load testing will tell us I guess.

john-herholz-dt: @sebalix can you tell where you set the option? We encounter a lot of "Socket is closed" errors unfortunatly.

JordiMForgeFlow: @john-herholz-dt simply choose a check connection method (either ls or marker file) and everytime it wants to connect it will check if connection needs to be reopened.

john-herholz-dt: @JordiBForgeFlow thanks for your answer! I checked again.
This PR was not forward ported to 17.0 branch that we were using. Any change to get this done quickly? I will check tomorrow if I can prepare a forward port myself but i am not so familiar with it.

JordiMForgeFlow: @john-herholz-dt all FWs were done. You can see it in v17: https://github.com/OCA/storage/blob/17.0/fs_storage/models/fs_storage.py#L149

john-herholz-dt: @JordiMForgeFlow okay you are right. I was confused because the def _check_connection method looks different but that is maybe due to different PR later.

So this brings me back to: We still have the issue with closed Socket and we have a test connection method selected for all FS Storage entries.

But I will further investigate it. Thanks for helping!

JordiMForgeFlow: @john-herholz-dt unfortunately I am not running this in any V17 project, so I cannot assure you that the changes they did in V17 are still handling correctly the issue.

Maybe you could try changing the _check_connection method back to how it is in V16 to see if it works

lmignon: @john-herholz-dt you could be interested by #487 In this PR I add the possibility to disable the cache of the fsspec client in odoo since fsspec already manage it's own cache. IMO, the cache into Odoo should be removed into the next version.

@JordiMForgeFlow @john-herholz-dt AFAIK 16, 17 and 18 branches are aligned. I regularly do back and forward ports between branches and every time I merge a PR. (But I'm not infallible...)

PR #321, opened 2024-01-12
[16.0][IMP] fs_product_multi_image: add labels on image fields

marielejeune:
Otherwise fields `image` and `image_medium` have the same labels as standard odoo fields.

### DIFF
--- a/fs_product_multi_image/models/product_product.py
+++ b/fs_product_multi_image/models/product_product.py
@@ -25,9 +25,14 @@ class ProductProduct(models.Model):
         # Store it to improve perfs
         store=True,
     )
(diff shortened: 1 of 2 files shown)


qgroulard (approved):

lmignon (approved):

lmignon:
@marielejeune @qgroulard #321 seems to be more complete.... Can you take a look?

marielejeune:
> @marielejeune @qgroulard #321 seems to be more complete.... Can you take a look?

#321 is the current PR...But if you want to speak about https://github.com/OCA/storage/pull/324, yes indeed, we can keep that one.

Issue #313, opened 2023-12-13
Documents app preview issue


JurajRech:
## Module
fs_attachment / fs_storage

## Describe the bug
When using s3 storage for all attachments - unable to preview file (pdf, jpg, gif .... ) in Documents app
![image](https://github.com/OCA/storage/assets/99398177/1f7ddb1f-327b-41b4-98ba-90fbf3323149)

**workaround**

if I upload an attachment (pdf) to TASK - I can preview it inside the task
but if I try to preview the PDF via Documents app, throws error like on screenshot
copy from docker container error:
```
2023-12-13 17:20:25,651 15 ERROR odoo odoo.http: Exception during request handling. 
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/odoo/http.py", line 1988, in __call__
    response = request._serve_db()
  File "/usr/lib/python3/dist-packages/odoo/http.py", line 1584, in _serve_db
    return service_model.retrying(self._serve_ir_http, self.env)
  File "/usr/lib/python3/dist-packages/odoo/service/model.py", line 134, in retrying
    result = func()
  File "/usr/lib/python3/dist-packages/odoo/http.py", line 1611, in _serve_ir_http
    response = self.dispatcher.dispatch(rule.endpoint, args)
  File "/usr/lib/python3/dist-packages/odoo/http.py", line 1724, in dispatch
    return self.request.registry['ir.http']._dispatch(endpoint)
  File "/usr/lib/python3/dist-packages/odoo/addons/base/models/ir_http.py", line 154, in _dispatch
    result = endpoint(**request.params)
  File "/usr/lib/python3/dist-packages/odoo/http.py", line 699, in route_wrapper
    result = endpoint(self, *args, **params_ok)
  File "/usr/lib/python3/dist-packages/odoo/addons/web/controllers/binary.py", line 75, in content_common
    stream = request.env['ir.binary']._get_stream_from(record, field, filename, filename_field, mimetype)
  File "/usr/lib/python3/dist-packages/odoo/addons/base/models/ir_binary.py", line 129, in _get_stream_from
    stream = self._record_to_stream(record, field_name)
  File "/mnt/extra-addons/documents/models/ir_binary.py", line 11, in _record_to_stream
    return Stream.from_attachment(record.attachment_id.sudo())
  File "/usr/lib/python3/dist-packages/odoo/http.py", line 475, in from_attachment
    stat = os.stat(self.path)
[...]

PR #322, opened 2024-01-12
[ADD][16.0] fs_document

gbrito:
Odoo with the document app introduces changes on how the document.document interacts with ir.attachment

This module will fix issue #313


asanjuan-kernet:
Had the exact same issue and couldn't download documents either, this fixed them both. Any updated on it?
Thanks for your work

fs_document/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+# Copyright (c) 2024 ERP|OPEN <https://www.erpopen.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+
+{
+    "name": "Base Document Attachment Object Store",
+    "summary": "Enable fetching external stored document.document",
+    "version": "16.0.1.0.0",
+    "author": "ERP|OPEN B.V. , Odoo Community Association (OCA)",
+    "license": "AGPL-3",
```
lmignon, line 10: This addon depends on an enterprise one. The `AGPL` licence is not compatible with the `OEEL-1` licence.

lmignon (commented):
Thank you for this usefull glue addon between `fs_attachment` and the `documents` addon from Odoo enterprise. Since the CI and our runboat infrastructure have no access to the Enterprise addons we must find a way to manage this kind of addons.

fs_document/__manifest__.py
gbrito: @lmignon maybe i can add the if statement on the base module. Ir the document module is not installed it will still work

PR #323, opened 2024-01-16
[MIG][17.0] fs_storage: Migration to 17.0

john-herholz-dt:


fs_storage/tests/common.py
```diff
@@ -16,7 +16,7 @@ class TestFSStorageCase(TransactionCase):
     def setUpClass(cls):
         super().setUpClass()
         cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
-        cls.backend: FSStorage = cls.env.ref("fs_storage.default_fs_storage")
```
john-herholz-dt: I changed this, as I think it is bad practice to get used with starting records with default_ as for example Odoo throws errors if you do this on fields.

Furthermore it is more precise, as this demo fs_storage is not "default", it is "demo".

gurneyalex (approved):

PR #325, opened 2024-01-17
[FIX]Adapt the storbinary to handle FTP cwd

flotho:
When using "complex" directory path, the command CWD place the client in a folder. Then when submitting the file, fullpath no more exits relatively to the folder where the client is placed. In this case the full_path has to be replace by the basenam so that no more additionnal "relative path" will be added when trying to storebinary


flotho:
actually it was the file name that caused the issue,
closing

PR #326, opened 2024-01-23
[16.0] fs_base_multi_image: widget to drag and drop images

benwillig:
New widget to easily create new relation image records with drag and drop.
File selection through browser dialog is also possible.

image


fs_product_multi_image/views/product_template.xml
```diff
@@ -20,7 +20,11 @@
             </field>
             <page name="sales" position="after">
                 <page name="images" string="Images">
-                    <field name="image_ids">
+                    <field
+                        name="image_ids"
+                        widget="fs_image_relation_dnd_upload"
+                        options="{'target': 'fs_image'}"
+                    >
```
lmignon: IMO, the target should be selected by the user and default on specific. Cases where you want to reuse the uploaded image for others objects are minority. You could for example add a select field into the DND zone on the widget to allows the selection of the target...

lmignon (changes requested):
@benwillig Thank you for this nice improvement! A little comment...

fs_product_multi_image/views/product_template.xml
benwillig: Yes I thought about this, I will add this.
Also I think it could be interesting to add a dialog where we can select existing fs.image, but this could be done afterwards.

fs_base_multi_image/__manifest__.py
```diff
@@ -18,6 +18,16 @@
         "views/fs_image.xml",
         "views/fs_image_relation_mixin.xml",
     ],
+    "assets": {
+        "web.assets_backend": [
+            "fs_base_multi_image/static/src/fields/"
+            "fs_image_relation_dnd_upload/fs_image_relation_dnd_upload.js",
```
lmignon: @benwillig To avoid the error into pre-commit, I thing you should rename your file to end with 'esm.js'.
```suggestion
            "fs_image_relation_dnd_upload/fs_image_relation_dnd_upload.esm.js",
```

benwillig: Indeed

fs_product_multi_image/views/product_template.xml
benwillig: I added the select to change the target.

lmignon:
@benwillig  I don't know if you encounter the same issue but sometime, I need to drag multiple times the same file to add the image.... Should we add a progress information while the creation is not yet completed? (Nice to have)

benwillig:
> @benwillig I don't know if you encounter the same issue but sometime, I need to drag multiple times the same file to add the image.... Should we add a progress information while the creation is not yet completed? (Nice to have)

I didn't experience this issue (either with firefox or chrome). But I guess we could simply block the user interface just after the file selection, what do you think?

lmignon (approved):
LGTM Works like a charm. (code review + functional tests)

gurneyalex (approved):

PR #331, opened 2024-01-29
[16.0] fs_storage: support SSH private keys authentication

sebalix:
SSH connections can now be done with private keys by setting the `pkey`+ `passphrase` options. Coupled with the `eval_options_from_env` this allows to set these ones from the environment, e.g:

`{"host": "sftp.example.net", "username": "odoo", "pkey": "$SSH_KEY", "passphrase": "$SSH_PASSPHRASE", "port": 22}`

### DIFF
--- a/fs_storage/models/fs_storage.py
+++ b/fs_storage/models/fs_storage.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import base64
 import functools
 import inspect
+import io
 import json
 import logging
 import os.path
@@ -20,6 +21,19 @@ from odoo.exceptions import ValidationError
 from odoo.addons.base_sparse_field.models.fields import Serialized
 
 _logger = logging.getLogger(__name__)
+
+try:
+    import paramiko
+
+    SSH_PKEYS = {
+        "DSS": paramiko.DSSKey,
+        "RSA": paramiko.RSAKey,
+        "ECDSA": paramiko.ECDSAKey,
+        "OPENSSH": paramiko.Ed25519Key,
+    }
+except ImportError:  # pragma: no cover
+    _logger.debug("Cannot `import paramiko`.")
+    SSH_PKEYS = {}
 
 
 # TODO: useful for the whole OCA?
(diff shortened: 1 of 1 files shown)


simahawk (commented):
LG. Could we have some test cov?

sebalix:
@simahawk what's the best approach here to test? Current module doesn't depend on `paramiko`, if this one is installed then `fsspec` can use it. To test the feature I would need to add `paramiko` as tests dependencies, is it OK?

Or should we put tests in a dedicated `test_fs_storage_ssh` module?

sebalix:
<s>I added an extra parameter `pkey_type` to force the key type (`RSA`, `DSS`, `OPENSSH`...).
It happens that some versions of `ssh-keygen` generates `OPENSSH` keys even with the `-t rsa` parameter if `-m PEM` is not set.
https://serverfault.com/questions/939909/ssh-keygen-does-not-create-rsa-private-key

Still need to test it a bit more.</s>

EDIT: this doesn't work neither. Better to generate the key in a proper manner with `ssh-keygen -t rsa -m PEM [...]`, and everything works as expected. I reverted my last changes.

simahawk:
> @simahawk what's the best approach here to test? Current module doesn't depend on `paramiko`, if this one is installed then `fsspec` can use it. To test the feature I would need to add `paramiko` as tests dependencies, is it OK?
> 
> Or should we put tests in a dedicated `test_fs_storage_ssh` module?

In fact, we are adding too many SFTP related feature to the base module... IMO we need a separated module for sftp: `fs_storage_sftp`.

florian-dacosta:
@sebalix Do you plan to continue and create a new module with the feature ?

FYI, I did manage a SFTP connection with a private key, using sshfs (https://github.com/fsspec/sshfs). If you install this lib, it will replace sftp default paramiko implementation by another based on asyncssh. 
This allow to manage keys out of the box (at least, it works in my case, without any additional code).

IMO, we could drop this and maybe write somewhere to install sshfs if you want to manage keys, or do you see any reason not to use the sshfs implementation ?

lmignon:
ping @sebalix @simahawk Do you've seen the last comment from @florian-dacosta? Can we close-it?

simahawk:
@sebalix ping

PR #332, opened 2024-01-29
[16.0][CHG] fs_base_multi_image: drag and drop fixes

benwillig:
- create in memory records instead of calling the ORM explicitly

### DIFF
--- a/fs_product_multi_image/views/product_template.xml
+++ b/fs_product_multi_image/views/product_template.xml
@@ -20,7 +20,11 @@
             </field>
             <page name="sales" position="after">
                 <page name="images" string="Images">
-                    <field name="image_ids" widget="fs_image_relation_dnd_upload">
+                    <field
+                        name="image_ids"
+                        widget="fs_image_relation_dnd_upload"
+                        mode="kanban"
+                    >
                         <tree>
                             <field name="sequence" widget="handle" />
                             <field name="name" />
--- a/fs_product_multi_image/views/product_template.xml
+++ b/fs_product_multi_image/views/product_template.xml
@@ -24,13 +24,12 @@
                         <tree>
                             <field name="sequence" widget="handle" />
                             <field name="name" />
+                            <field name="link_existing" invisible="1" />
+                            <field name="image_id" invisible="1" />
+                            <field name="image" invisible="1" />
+                            <field name="specific_image" invisible="1" />
                         </tree>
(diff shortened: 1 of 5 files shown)


lmignon:
@benwillig Why do we need to create the relation on drag and drop? Can we add a relation and save it when the parent is saved?

benwillig:
@lmignon That's what I wanted to do, unfortunately I didn't find the right way to do it with the new framework.

benwillig:
@lmignon I finally found the way to make it work as it was in the old widget. And it also works when saving a specific image directly on the relation record.

lmignon (approved):
Great! Works like a charm.

lmignon:
@bealdav Can you take a look at this?

bealdav:
Yes I 'll dive in.

bealdav:
Thanks for this improvement even if don't understand the full details.

Nice to see Kanban view completed with some fields.

As product.product view doesn't allow to edit images, don't you think that this view could show kanban view instead of tree ? It's a better visual view for users.

benwillig:
Yes we can display the kanban view by default. What do you think @lmignon ?

lmignon:
> Yes we can display the kanban view by default. What do you think @lmignon ?

Yes we can on the product.product view.

bealdav (approved):
Thanks a lot. LGTM
Kanban in product.product view is a  nice to have  I think

bealdav:
It works nicely in prod db, really thanks

bealdav:
Also this PR https://github.com/OCA/storage/pull/370 seems ok. What do you think ?

fs_product_multi_image/views/product_template.xml
```diff
@@ -20,17 +20,20 @@
...
+                        mode="kanban"
+                    >
                         <tree>
                             <field name="sequence" widget="handle" />
                             <field name="name" />
+                            <field name="link_existing" invisible="1" />
+                            <field name="image_id" invisible="1" />
+                            <field name="image" invisible="1" />
+                            <field name="specific_image" invisible="1" />
                         </tree>
```
lmignon, line 35: Is-it the right way to define the kanban?

benwillig: I added a specific kanban view for the image relation model so we don't need to define an inline kanban view here (if this is what you asked).

PR #336, opened 2024-02-07
[16.0][FIX] fs_storage: fix test connection when __fs slot is not initialized

JordiMForgeFlow:
When the __fs slot has not yet been initialized the test config method fails. As the check connection is called when accessing the property, we can simply access the property to test. Comes after https://github.com/OCA/storage/pull/320

CC @lmignon

### DIFF
--- a/fs_storage/models/fs_storage.py
+++ b/fs_storage/models/fs_storage.py
@@ -454,7 +454,8 @@ class FSStorage(models.Model):
 
     def action_test_config(self) -> None:
         try:
-            self._check_connection(self.__fs)
+            # pylint: disable=W0104
+            self.fs
             title = _("Connection Test Succeeded!")
             message = _("Everything seems properly set up!")
             msg_type = "success"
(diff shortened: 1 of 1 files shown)


fs_storage/models/fs_storage.py
```diff
@@ -454,7 +454,8 @@ def delete(self, relative_path) -> None:
 
     def action_test_config(self) -> None:
         try:
-            self._check_connection(self.__fs)
+            # pylint: disable=W0104
+            self.fs
```
sebalix, line 460: A comment is welcome

sebalix (approved):

fs_storage/models/fs_storage.py
JordiMForgeFlow: @sebalix I have added a change to clear the slot when the connection fails, maybe this fixes the issue where the FS object was being kept although the connection was closed

florian-dacosta (approved):

lmignon (approved):

PR #347, opened 2024-03-01
[14.0][IMP] Added the capability to filter on sftp backend

wouitmil:
Those change allows to filter file list with regex pattern but also on file type (folder/file/link)


santostelmo (approved):

santostelmo:
@wouitmil code looks good. But tests are falling

simahawk:
@wouitmil tests are red. Also please, drop the Odoo version from the commit msg

PR #349, opened 2024-03-05
16.0: issue with fs_attachment. Temporary hack for s3

sebastienbeau:
[DO NOT MERGE, it's an hack]

@lmignon @bealdav I have the following issue when using S3 bucket from scaleway.

In the case of scaleway when you when to make your image public you have to configure the acl "public-read" 
This work perfectly when uploading a file but when doing a rename, the lib will do a copy and rm. And the acl is not transferred.

The image are always private and never public.

This is an temporary hack, I share it to help other, but I think the issue is more in s3fs lib.

Note: I didn't had the time to do a real analyse but renaming have a cost for some provider (not big) https://stackoverflow.com/questions/33098211/does-rename-a-file-cost-money-on-s3. So maybe it's better to find a solution to directly push the file with the right name.

### DIFF
--- a/fs_attachment/models/ir_attachment.py
+++ b/fs_attachment/models/ir_attachment.py
@@ -465,7 +465,9 @@ class IrAttachment(models.Model):
             # rename on S3 is a copy with an rm
             # and rename do not transfert acl correctly
             # so we need to add a chmod after the rename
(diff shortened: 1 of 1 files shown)


sebastienbeau:
I close the PR, after looking twice the acl is a legacy feature and bucket policy should be used instead and with policy this hack is not needed

PR #354, opened 2024-03-15
[16.0] fs_attachment: sent the mimetype information for S3

sebastienbeau:
When storing image on amazon s3, ovh or scaleway.
The minetype must be set in order to have the right mimetype when requesting the image publicly.

Note : when using the api put_file it will guess the mimetype automatically : https://github.com/fsspec/s3fs/blob/47ffcfad33f8c74974a1515cb266168ede6cf4b5/s3fs/core.py#L1176

### DIFF
--- a/fs_attachment/models/ir_attachment.py
+++ b/fs_attachment/models/ir_attachment.py
@@ -242,7 +242,9 @@ class IrAttachment(models.Model):
                     "db_datas": data,
                 }
                 return values
-        return super()._get_datas_related_values(data, mimetype)
+        return super(
+            IrAttachment, self.with_context(mimetype=mimetype)
+        )._get_datas_related_values(data, mimetype)
 
     ###########################################################
     # Odoo methods that we override to use the object storage #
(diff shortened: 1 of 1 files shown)


vincent-hatakeyama:
Cherry picked in 17.0 to test with a local minio, mimetype is now correctly set.

fs_attachment/models/ir_attachment.py
```diff
@@ -381,6 +383,13 @@ def _storage_file_read(self, fname: str) -> bytes | None:
             )
         return b""
 
+    def _storage_write_option(self, fs):
+        if hasattr(fs.fs, "s3"):
+            # For S3 storage the mimetype should be pass in the kwargs
+            return {"ContentType": self._context["mimetype"]}
```
lmignon, line 389: the s3 fs could be nested at a deeper level... IMO you should recursively test the nested fs. Something like:

```
_fs = fs
while _fs:
    if hasattr(_fs, "s3"):
        return {"ContentType": self._context["mimetype"]}
    _fs = getattr(_fs, "fs", None)
```

lmignon (changes requested):

lmignon:
superseded by #378

sbidoul:
heads-up, in #497 this is moving to a separate fs_attachment_s3 module

PR #355, opened 2024-03-15
[16.0][FIX] fs_storage: remove default protocol

sbidoul:
Since protocol is under server_environment control, it seems its default value takes priority over a value set in an XML record.

Since it does not really make much sense to create a storage backend without specifying the protocol, we remove the default value.


adrienpeiffer (approved):

sbidoul:
@lmignon ok with this?

lmignon (approved):

PR #357, opened 2024-04-05
[16] fs_storage : Allow to choose different method to check connection

florian-dacosta:
Since https://github.com/OCA/storage/pull/320,  in order to resolve some cache issues, the way to check connection did change and became mandatory (which was not the case before)

Now, a marker file is created on remote server which is then read each time we need to connect.
I did ran into some issues because of this on server with limited access rights.. (Not allowed to write in the remote server for instance).
Another issue is that some partners don't expect Odoo to send these kind of files (the marker files) and their system is not able to ignore it for instance.

I find this way quite intrusive, when connecting to remote server that are managed by external partners.
We could have to deal with a server with read-only access, and this is not possible anymore.

To solve this issue, I propose to give an option on the fs storage on how the connection should be checked.
By default, the mark file stuff is set, but you could choose or implement other methods it this one does not work for your case.

I did implement the one we had before (list file on the whole folder), but I think it is valid to implement still other methods if necessary.

@hparfr @lmignon @JordiMForgeFlow @sebalix @nilshamerlinck


JordiMForgeFlow:
For me it looks good to have the option to decide, I would just maybe recommend to add some description in the README on the implications of using each method.

florian-dacosta:
I am not sure we want to go into such detail in the README, but at least a help on the field to describe each option would be nice indeed

fs_storage/models/fs_storage.py
```diff
@@ -148,6 +148,14 @@ def __init__(self, env, ids=(), prefetch_ids=()):
...
         store=False,
     )
+    check_connection_method = fields.Selection(
+        [
+            ("marker_file", "Create Marker file"),
+            ("ls", "List File"),
+        ],
+        default="marker_file",
+        required=True,
+    )
```
lmignon, line 158: A user friendly description would be welcome.... Moreover I would make it not required...

lmignon (changes requested):
Thank you for this improvment @florian-dacosta . Can you add a paragraph into the `usage.rst` file to explain the check connection mechanism and when to use one or an other.

fs_storage/models/fs_storage.py
florian-dacosta: I put it required because we need a method to check the connection anyway. As there is a default value, it did not seems to be a problem to me.
If we do not put it as required, it means we go with the marker file method if empty and it just seems more hidden for the user.
On my side I don't really care actually

lmignon: A non required field will allow to ignore the check mechanism...

florian-dacosta: That is a good idea. Before the cache issue, the check was not done (it was only with manual button action_test_config) and it was not an issue in my use cases. The fact to check the connection before beeing able to interact with remote server is actually a new "cost" (in time) and it can be nice to make it optional indeed.

The only issue I see, is that in case we set it empty, the manual button `action_test_config` won't work either.
I think it is a shame, you may want to avoid the check mechanism each time you interact with the remote server but still be able to easily check all is fine manually.
Any idea to manage that ? We could force a method in this case but which ?... I'd go with the `ls` one which is less intrusive IMO but not sure it will be ok for every body.

I'll update the PR this way trying to be as clear as I can in the README about the impact of this field.
I guess I'll let the default in order to stay consistent with current behvior.

lmignon: @florian-dacosta The check button could open a wizard with the same selection field but required....

What we could do:

* define a method on the model to get the possible values for the `check_connection_method`
* modify the  `check_connection_method` field to get possible values from the new method
* define a new wizard to use to test the connection with a `check_connection_method` method field required that will get the possible value from the method on the `fs.storage` model
* replace the logic behind the `action_test_config` method to return a wizard...

What do you thing about?

florian-dacosta:
> For me it looks good to have the option to decide, I would just maybe recommend to add some description in the README on the implications of using each method.

Done @JordiMForgeFlow

fs_storage/wizards/fs_test_connection.py
```diff
@@ -0,0 +1,28 @@
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from odoo import api, fields, models
+
+
+class FSTestConnection(models.Model):
```
JordiMForgeFlow: should be transient model

florian-dacosta: Oops indeed.
It is fixed

fs_storage/wizards/fs_test_connection.py
```diff
@@ -0,0 +1,28 @@
...
+    @api.model
+    def default_get(self, field_list):
+        res = super().default_get(field_list)
+        res["storage_id"] = self.env.context.get("active_id", False)
+        return res
+
+    def action_test_config(self):
+        return self.storage_id.with_context(
+            force_connection_method=self.check_connection_method
+        )._test_config()
```
lmignon: Why do you pass the forced method into the context? It could be an optional argument of the `test_config` method...

```suggestion
        return self.storage_id._test_config(force_check_method=self.check_connection_method)
```

JordiMForgeFlow (approved):

fs_storage/models/fs_storage.py
```diff
@@ -268,14 +284,29 @@ def _compute_options_properties(self) -> None:
...
+            self._marker_file_check_connection(fs)
+        elif check_connection_method == "ls":
+            self._ls_check_connection(fs)
         return True
 
+    def _check_connection(self, fs):
+        check_connection_method = self.check_connection_method or self.env.context.get(
+            "force_connection_method", ""
+        )
+        return self._check_connection_with_method(fs, check_connection_method)
```
lmignon: IMP We don't need to pass args via context
```suggestion
        return self._check_connection_with_method(fs, self.check_connection_method)
```

lmignon (changes requested):

fs_storage/wizards/fs_test_connection.py
florian-dacosta: It is because I did not want to pass an arg to `fs` property method. And connection is checked by calling the fs property...
And I did not especially want to change the way connection is checked.

fs_storage/models/fs_storage.py
florian-dacosta: The same method is used when connection is checked from the wizard, for a `fs.storage` with empty `check_connection_method`

fs_storage/wizards/fs_test_connection.py
lmignon: but you could also call
``` python
self.storage_id._check_connection_with_method(self.storage_id.fs, self.check_connection_method)
```
The same effect and no need to use the context to pass an args... I prefer explicit way to pass args...

fs_storage/models/fs_storage.py
lmignon: We don't need to use the context to pass args from wizard. We can directly call the `_check_connection_with_method` from the wizard.

fs_storage/wizards/fs_test_connection.py
florian-dacosta: Well yes, I could, this will then call `_check_connection_with_method` twice actually : 
Once because of `self.storage_id.fs`, which will call the `_check_connection_with_method` and the second time because we call it directly.
I don't want to duplicate the code of `_test_config` so it behave differently in case it is called from wizard or from a fs.storage with a `check_connection_method`configured...

Then I see 2 options if we really get rid of the context : 
1. Always call `_check_connection_with_method` in case of manual check, it will check twice if the fs.storage is configured to have a method but it is no big deal since it only happens on manual action (which is rarely done)
2. When checking the connection manually, we always open the wizard, even if the `check_connection_method` is set.
I would rather not, because it adds an extra step and it give the choice to choose a different `check_connection_method` in the wizard than the one configured on the `fs.storage`...

I also prefer explicit way to pass args, but it seemed like the easiest way to avoid too much code duplication/addition.

Anyway, I am not against the 1 I did a commit toward this way.

fs_storage/models/fs_storage.py
florian-dacosta: The call on `_test_config` instead of `_check_connection_with_method` from the wizard is there to use the same notification mechanism.

lmignon (approved):
Thank you for this great improvement. LGTM (Code review)

lmignon:
@JordiMForgeFlow can you take a look at last changes?

JordiMForgeFlow (approved):
It looks good to me 👍🏼

PR #360, opened 2024-04-16
[16.0][MIG-REF] storage_import_image_advanced -> fs_import_image_advanced, fs_import_image_advanced_thumbnail

ChrisOForgeFlow:
- Refactor from storage_import_image_advanced
- Change to use fs_storage and fs_product_multi_image as dependencies
- Fix test to use properly test class dependency
- Removed storage_id field, setup of model or fields is unique way to address fs in new model
- Removed thumbnail support, because module is not required by module fs_product_multi_image, added a new glue module 'fs_import_image_advanced_thumbnail' support to create automatic thumbnails when module fs_image_thumbnail is installed


AlexPForgeFlow:
can we reopen?

PR #361, opened 2024-04-22
[16.0][IMP] fs_attachment: No crash on missing file

lmignon:
When trying to get access to files stored into an external filesystem, don't crash if the files cannot be accessed. This is necessary to avoid blocking Odoo's operation if an external system is not accessible.


CasVissers-360ERP (approved):
Functional review

PR #364, opened 2024-04-26
IMP fs_product_multi_image: allow to also display kanban in Images tab

bealdav:
We need to discuss the way we need/can display images.

Here is a dirty way to have kanban and drag and drop

image


lmignon:
@bealdav A nice improvement would be for the widget to take into account the mode attribute to know how to display the images. Depending on project needs, you could simply override the view to change the mode.

PR #365, opened 2024-04-27
[16.0][FIX] fs_attachment: respect controller parameters for /web/content/

hbrunn:
when we craft a link like

`/web/content?model=my.model&filename_field=filename&field=binary_field&mimetype=mime.type&filename=myfile.name`

the parameters are ignored and the resulting download gets the generic (wrong) name/mimetype from the attachment of the binary field if it is of type `fs`. For filestore attachments, this happens in [`ir.binary#_get_stream_from`](https://github.com/OCA/OCB/blob/16.0/odoo/addons/base/models/ir_binary.py#L131), so I think we need to do the same thing for fs attachments. Otherwise, we can't really use this as drop-in replacement for storing binary fields in object storage.

### DIFF
--- a/fs_attachment/models/ir_binary.py
+++ b/fs_attachment/models/ir_binary.py
@@ -44,6 +44,34 @@ class IrBinary(models.AbstractModel):
             return FsStream.from_fs_attachment(fs_attachment)
         return super()._record_to_stream(record, field_name)
 
+    def _get_stream_from(
+        self,
+        record,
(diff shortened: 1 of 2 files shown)


lmignon:
Thank you for the fix @hbrunn Can you I a unittest to check that it works in the initial case and with fs_attachment?

lmignon (approved):

PR #366, opened 2024-04-27
[16.0][FIX] fs_attachment: Rename file in storage when name is written

hbrunn:


lmignon:
Thank you for this new fix @hbrunn. Can you add a unittest? Thank you  in advance.

lmignon (approved):

PR #370, opened 2024-05-14
[IMP] fs_product_multi_img: add filter on search product view

bealdav:
Please could you have a look ?

cc @benwillig @hparfr @florian-dacosta @HviorForgeFlow

### DIFF
--- a/fs_product_multi_image/views/product_template.xml
+++ b/fs_product_multi_image/views/product_template.xml
@@ -47,4 +47,23 @@
             </xpath>
         </field>
     </record>
+    <record id="product_template_search_view" model="ir.ui.view">
+        <field name="model">product.template</field>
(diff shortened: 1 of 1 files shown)


benwillig (commented):
LGTM. But I'm not sure those filters would be relevant for all instances so I don't know if we should add them here. I guess those could be personal user filters.

bealdav:
> But I'm not sure those filters would be relevant for all instances

OK, you're probably right. On my side I have to find a way to replicate this kind of customization on several projects. Personal user filters is a really bad experience because those are not cumulative filters, migration is not so easy and replication over projects is impossible.

Then I plan to make an additional module for these kind of usability features. Then probably in Akretion organization.

Thanks for your review and feedback.

PR #371, opened 2024-05-16
PIN fsspec requirement because change in make_path_posix() signature

bealdav:
https://github.com/fsspec/filesystem_spec/pull/1567/files#diff-b47117e431a1b37790be709fb84c0d854da5e73203e3de57ec6b6032b18eebf1R280

spec

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
 # generated from manifests external_dependencies
-fsspec
+fsspec==2024.3.1
 paramiko
 python_slugify


bealdav:
cc @lmignon @benwillig

lmignon (changes requested):
@bealdav What's the problem?  I would prefer to adapt the code to the new version than being locked to a specific version. Moreover, the version constrain should be specified into the manifest file.

bealdav:
Yes you're right. This PR is just a quick alert for community

lmignon:
@bealdav can you take a look at https://github.com/OCA/storage/pull/372

bealdav:
Thanks to warned me, perfect

PR #372, opened 2024-05-17
[16.0][FIX] fs_storage: Change make_path_posix() to only 1 argument

victoralmau:
Change `make_path_posix()` to only 1 argument

Since https://github.com/fsspec/filesystem_spec/commit/da7754870bb9ab30c99fcd6dccdd39421f420691 it is necessary to set in `make_path_posix()` only 1 argument

@Tecnativa

### DIFF
--- a/fs_storage/__manifest__.py
+++ b/fs_storage/__manifest__.py
@@ -19,5 +19,5 @@
         "wizards/fs_test_connection.xml",
     ],
     "demo": ["demo/fs_storage.xml"],
-    "external_dependencies": {"python": ["fsspec"]},
+    "external_dependencies": {"python": ["fsspec>=2024.5.0"]},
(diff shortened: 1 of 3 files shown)


pedrobaeza:
@lmignon do you prefer this solution or to pin the library version?

lmignon (changes requested):
Thank you for the change.

Can you also modify the manifest file to specify the minimal version for the fsspec python dependency as:

```python
 "external_dependencies": {"python": ["fsspec>=2024.5.0"]},
```

Thank you.

victoralmau:
Changes done.

lmignon (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #375, opened 2024-06-04
[14.0][FIX] storage_backend_ftp: Fix the list method in ftp_adapter.py to return file names instead of full paths.

CedricAtanane:
This change aligns the behavior with s3_adapter. 
Detected when using edi_oca_storage, files were not being moved.


lmignon (approved):
LGTM (Code review only)

PR #379, opened 2024-06-24
[17.0][IMP] fs_storage: replace {db_name} in directory_path, fs_attachment and fs_attachment_s3: adapt to change

vincent-hatakeyama:
In multi database mode, there was no way to avoid filename collisions.

In multi database mode, without using that feature in fs_attachment, as the created filename does not contain the database name, it is possible to overwrite files from another database if obfuscation is off and the file have the same name and the same id or cause deletion of files if the checksum is used as the name and the same file is copied over several database and unlinked in one of them.
I would have prefer for the name of the file to always include the database name but I think it would have been a major change considered too breaking. Allowing to use the database name seemed more compatible.

### DIFF
--- a/fs_storage/models/fs_storage.py
+++ b/fs_storage/models/fs_storage.py
@@ -124,8 +124,10 @@ class FSStorage(models.Model):
             """,
     )
 
+    # When accessing this field, use the method get_directory_path instead so that
+    # parameter expansion is done.
     directory_path = fields.Char(
-        help="Relative path to the directory to store the file"
+        help="Relative path to the directory to store the file",
     )
 
     # the next fields are used to display documentation to help the user
@@ -187,6 +189,14 @@ class FSStorage(models.Model):
         self.__fs = None
         self.env.registry.clear_cache()
         return super().write(vals)
+
+    def get_directory_path(self):
+        """Returns directory path with substitution done."""
+        return (
+            self.directory_path.format(db_name=self.env.cr.dbname)
+            if isinstance(self.directory_path, str)
+            else self.directory_path
+        )
 
     @api.model
     @tools.ormcache()
@@ -409,7 +419,7 @@ class FSStorage(models.Model):
             options["auth"] = tuple(options["auth"])
         options = self._recursive_add_odoo_storage_path(options)
         fs = fsspec.filesystem(self.protocol, **options)
-        directory_path = self.directory_path
+        directory_path = self.get_directory_path()
         if directory_path:
             fs = fsspec.filesystem("rooted_dir", path=directory_path, fs=fs)
         return fs
--- a/fs_storage/readme/USAGE.md
+++ b/fs_storage/readme/USAGE.md
@@ -75,6 +75,10 @@ for the following fields:
 - options
 - directory_path
 
+When evaluating directory_path, `{db_name}` is replaced by the database name.
+This is usefull in multi-tenant with a setup completly controlled by
+configuration files.
+
 ## Migration from storage_backend
 
 The fs_storage addon can be used to replace the storage_backend addon.
--- a/fs_attachment/models/fs_storage.py
+++ b/fs_attachment/models/fs_storage.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 
 from odoo import _, api, fields, models, tools
 from odoo.exceptions import ValidationError
+from odoo.tools import SQL
 from odoo.tools.safe_eval import const_eval
 
 from .ir_attachment import IrAttachment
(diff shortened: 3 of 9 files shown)


lmignon:
@vincent-hatakeyama you should use one backend by database. IMO this change is useless.

sbidoul (changes requested):
The use case is unclear to me and this could indeed be solved with a different backend configuration per database.

I can imagine this could facilitate the automatic creation of that backend.

This would need a test, and an update of the help text of the field to reveal the feature to users.

fs_storage/models/fs_storage.py
```diff
@@ -162,6 +162,13 @@ def __init__(self, env, ids=(), prefetch_ids=()):
     def _server_env_fields(self):
         return {"protocol": {}, "options": {}, "directory_path": {}}
 
+    def _server_env_read_from_config(self, field_name, config_getter):
+        value = super()._server_env_read_from_config(field_name, config_getter)
+        # replace {db_name} with the dbname
+        if field_name == "directory_path":
+            value = value.format(db_name=self.env.cr.dbname)
+        return value
```
sbidoul: This is also probably not the best place to implement this, as if we do this, it should not behave differently when the value is entered manually or via server environment.

vincent-hatakeyama:
To clarify the use case:

- Odoo is used in multi-tenant mode. New database can be created by some users.
- fs_storage and fs_attachment are used
- the configuration to store the files is stored in odoo configuration file.
- The files referenced by a database should not be deleted when deleted in another base.

With those point in mind, I thought that allowing db_name and changing its value would allow splitting the file per database.

> you should use one backend by database. IMO this change is useless.

In the case I’ve described, this is not possible. To my knowledge, it is not possible to indicate to server_environment to use db_name in any of the fields.

sbidoul:
Thanks for the clarification. This makes sense.

adrienpeiffer:
@vincent-hatakeyama  Thanks for that
Indeed, it makes sense. However, as Stéphane said, it would be preferable to put this in another place so that it's also possible when the value is set manually.

vincent-hatakeyama:
I plan to change how this is handled when I have the time and to add tests to cover the case.

khalidchamakh:
hello @vincent-hatakeyama, 
Any news about your PR? We're having the same problem.

vincent-hatakeyama:
The current version in the PR is working and is in running in productions (for at least a year).

I haven’t had the time yet to rework the PR.

Lionel-cl:
I've been working on this topic. I'll post my proposal today in a new PR.

vincent-hatakeyama:
I’ve reworked the PR.

Substitution of `{db_name}` is now done in all cases.
I choose to introduce a get_directory_path method to handle the substitution. That seemed easier than introducing another field and make directory_path computed while also configurable with server_environment.
While looking at it, I found out that adaptations are also needed in fs_attachment.

Side note, the module uses {db_name} to be similar with the substitution done in web_ribbon.

fs_storage/models/fs_storage.py
vincent-hatakeyama: Reworked as indicated in https://github.com/OCA/storage/pull/379#issuecomment-2970362090

fs_storage/readme/USAGE.md
```diff
@@ -75,6 +75,8 @@ for the following fields:
 - options
 - directory_path
 
+When evaluating directory_path, `{db_name}` is replaced by the database name.
```
lmignon, line 78: @vincent-hatakeyama Can you explain the use case where it could be a good practice to use the `{db_name}` placeholder in your directory path?

vincent-hatakeyama: I’ve added a sentence to indicate when the feature is to be used and rebased the branch.

lmignon (approved):
Thank you @vincent-hatakeyama for your contribution, your patience and your tenacity so that we understand the need ;-)

Lionel-cl:
Thanks for that. My PR is then no longer necessary. 
However, this code (although more conventional) also requires changing all modules dependent on fs_storage.

vincent-hatakeyama:
@lmignon Any chance this gets merged as the port to 18.0 (#514) has been merged?

lmignon (changes requested):
@vincent-hatakeyama  A fix seems to be missing for `fs_attachment_s3` https://github.com/OCA/storage/pull/514/files#diff-f384ce2e99b19f6cb9bd99fa6a15fa155805bcfb90bf5861fc90a146305fdf19

lmignon (approved):

PR #381, opened 2024-07-02
[16.0][CHG] fs_product_multi_image: fs.product.category.image views

benwillig:


fs_product_multi_image/views/product_category.xml
```diff
@@ -7,6 +7,7 @@
         <field name="name">product.category.form</field>
         <field name="model">product.category</field>
         <field name="inherit_id" ref="product.product_category_form_view" />
+        <field name="priority" eval="99" />
```
lmignon, line 10: Why do you put a specific priority. This could lead to error if others views with lower priority use the image_ids field as anchor to add new fields.

benwillig: Then I'm almost sure the image field is at the bottom of the form. If a view target the image_ids field, it should inherit from the view which add the field, not the base view.

lmignon (approved):

PR #382, opened 2024-07-03
[17.0][IMP] fs_storage: Allow setting eval_options_from_env in configuration file

vincent-hatakeyama:
eval_options_from_env can not be set in the configuration file, which prevents using environment variables to avoid displaying the secret key in the UI.


sbidoul:
Why not, but I have personally not have had problems with this since we set `eval_options_from_env` when creating the record.

lmignon (approved):

PR #383, opened 2024-07-09
[14.0][FIX] storage_thumbnail: always use slugified url_key

sebalix:
Both for creating and searching thumbnails.

This avoid an infinite creation of thumbnails on images as soon as an `url_key` is provided.

This is a regression introduced by commit 197659721718024b520c2ff364bd97b15dfc0b4b

Ref. 4622

### DIFF
--- a/storage_thumbnail/tests/test_thumbnail.py
+++ b/storage_thumbnail/tests/test_thumbnail.py
@@ -51,6 +51,18 @@ class TestStorageThumbnail(SavepointComponentCase):
         thumb.unlink()
         self.assertTrue(file_id.to_delete)
 
+    def test_image_get_or_create_thumbnail_no_duplicate(self):
+        # Ensure no duplicate is generated thanks to unique url_key
+        image = self._create_image()
+        self.assertTrue(image.url)
+        self.assertEqual(len(image.thumbnail_ids), 2)
+        thumb = image.thumb_small_id
+        thumb.url_key = "test"
+        existing_thumb = image.get_or_create_thumbnail(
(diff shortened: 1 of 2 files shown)


storage_thumbnail/tests/test_thumbnail.py
```diff
@@ -51,6 +51,17 @@ def test_thumbnail(self):
         thumb.unlink()
         self.assertTrue(file_id.to_delete)
 
+    def test_image_get_or_create_thumbnail(self):
+        image = self._create_image()
+        self.assertTrue(image.url)
+        self.assertEqual(2, len(image.thumbnail_ids))
```
simahawk:
```suggestion
        self.assertEqual(len(image.thumbnail_ids), 2)
```

storage_thumbnail/tests/test_thumbnail.py
```diff
@@ -51,6 +51,17 @@ def test_thumbnail(self):
         thumb.unlink()
         self.assertTrue(file_id.to_delete)
 
+    def test_image_get_or_create_thumbnail(self):
+        image = self._create_image()
```
simahawk:
```suggestion
    def test_image_get_or_create_thumbnail_no_duplicate(self):
        # Ensure no duplicate is generated thanks to unique url_key
        image = self._create_image()
```

simahawk (approved):
LG.
@sebastienbeau @lmignon warning for your existing instance.
We've got tons of duplicated thumbs after pulling in https://github.com/OCA/storage/pull/268 :wink:

lmignon (approved):
Sorry for the side effect introduced when merging #268

PR #384, opened 2024-07-11
[17.0][IMP] fs_storage: invalidate orm cache when connection fails

### LINKED ISSUE #374: [17.0] fs_storage: Connection closed after some time
Hello!

## Module: fs_storage

I think my issue is related to #319, but for V17.

I set up a FTP fs_storage, and it works great except that after some time (can be 10 seconds or a few minutes), it is impossible to communicate with the filesystem
When downloading:
image
When uploading:
image

If I restart my Odoo instance, it works again.

Like #319, I think the file system expires but is not renewed by Odoo.

[Here](https://github.com/OCA/storage/blob/962393fe2aabc738f1f4a42a614621e0d6c77654/fs_storage/models/fs_storage.py#L197), the filestore is saved in cache by Odoo. 
When I remove the `ormcache` decorator, the filesystem is created every time, and I don't have any problems.

As mentionned in #319 and #320, with heavy traffic, it is not recommended to fetch the filesystem every time. As it is not my case, I will remove the `ormcache` decorator while looking for a permanent solution.

Algosse:
Fixes #374 

I adapted a some parts of the code related to [_check_connection](https://github.com/OCA/storage/blob/3c19f00c9b1825afbf5a629e192309a9344e2e04/fs_storage/models/fs_storage.py#L297) from V16.0 (started in #320) and implemented it in V17.0.

Also, I added a check_connection everytime the fs is accessed to detected if it should be generated again or not.

### DIFF
--- a/fs_attachment/models/ir_attachment.py
+++ b/fs_attachment/models/ir_attachment.py
@@ -481,7 +481,7 @@ class IrAttachment(models.Model):
         code: str,
     ) -> fsspec.AbstractFileSystem | None:
         """Return the filesystem for the given storage code"""
-        fs = self.env["fs.storage"].get_fs_by_code(code)
+        fs = self.env["fs.storage"].sudo().get_fs_by_code(code)
         if not fs:
             raise SystemError(f"No Filesystem storage for code {code}")
         return fs
--- a/fs_storage/models/fs_storage.py
+++ b/fs_storage/models/fs_storage.py
@@ -194,8 +194,28 @@ class FSStorage(models.Model):
         return [s.code for s in self.search([])]
 
     @api.model
-    @tools.ormcache("code")
     def get_fs_by_code(self, code):
+        """Return the filesystem associated to the given code.
+
+        :param code: the code of the filesystem
+        """
+        fs = self._get_fs_by_code(code)
+        try:
+            self._check_connection(fs)
+        except Exception as e:
+            _logger.warning(
+                "Error while connecting to the filesystem storage %s: %s",
+                code,
+                e,
+            )
+            # Generate a new fs instance
+            self.env.registry.clear_cache()
+            fs = self._get_fs_by_code(code)
+        return fs
+
+    @api.model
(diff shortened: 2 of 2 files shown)


lmignon:
Thank you @Algosse for all your hard work. I don't work with Odoo version 17. It is therefore difficult for me to give an in-depth review of the proposed changes. I will try to take the time to install a version 17 and understand the problem you are experiencing.

sbidoul:
I have ported all missing commits from 16 to 17 in https://github.com/OCA/storage/pull/397

@Algosse  Could you check if that branch works for you?

lmignon:
@Algosse Is your fix still required with the last changes on 16.0?

Algosse:
Hello Guys,

Thanks for the forward of V16!

Sorry for the delay, I did not find the time to properly check and run the changes. 

I'm not sure the forward will solve my issue. In my previous tests, I figured the problem was the ormcache of the [get_fs_by_code method](https://github.com/OCA/storage/blob/0f35e5c3d1c94e2c56a7d9a623b7b1e8c12fb9a7/fs_storage/models/fs_storage.py#L217)

If the connection fails, the method keeps returning the same faulty fs. It does not access the fs property so it does not triger the check_connection method.
That is why I updated the get_fs_by_code method so it performs a check everytime (I know it can be costly but it is the best way I found yet). To get the fs, this new method calls the old one, renamed into a private method _get_fs_by_code.


I don't have the time to look at it now, I will be able in a few weeks. 

If someone wants to try, the "best" way I found to crash the connection was to use a fs_storage with SFTP protocol, linked it to a model (for example sale.order) and drag & drop multiple files at once in the chatter (I think Odoo tries to upload files in parallel and fsspec does not like it, at least for SFTP).

john-herholz-dt:
@Algosse what is the state of this issue for you? I am facing the same problem. SFTP Connections fail after multiple files.
I have updated latest 17.0 branch but it was not resolved for me.

PR #386, opened 2024-07-17
[16.0][FIX] fs_attachment: Prevent recompute_urls from skipping records with res_field != False

PabloEForgeFlow:
Currently `recompute_urls` will skip records which have a non-empty `res_field`, resulting in records with a valid `fs_storage_id` not being updated. The comment and fix were directly copied from [_force_storage_to_object_storage](https://github.com/OCA/storage/blob/ca197b1b2c12ec7e98915ead6b3d09b86a3a9efd/fs_attachment/models/ir_attachment.py#L719).

### DIFF
--- a/fs_attachment/models/fs_storage.py
+++ b/fs_attachment/models/fs_storage.py
@@ -455,8 +455,18 @@ class FsStorage(models.Model):
         in staging are done in a different directory and will  not impact the
         production.
         """
-        attachments = self.env["ir.attachment"].search(
-            [("fs_storage_id", "in", self.ids)]
-        )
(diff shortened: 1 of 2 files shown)


lmignon (approved):
Thank you for the fix @PabloEForgeFlow 

A nice to have should be to have a unittest fot this method

PabloEForgeFlow:
Sorry for the delay, I was a bit busy last week. I have added a test for `recompute_urls`, let me know what you think.

lmignon (approved):

BernatPForgeFlow (approved):
Code review looks good!

PR #404, opened 2024-10-10
[16.0][FIX] fs_attachment: fix fs_url computing

kobros-tech:
- Assuming that the optimized directory is checked and the directory path is not checked:
    - If the image is checked as obfuscated: 
        - then the fs_url refers to the image in its/dir1/dir2 directories that are named from checksum first 4 digits and this is fine.

    - But, if the image is not obfuscated:
        - then the "_get_url_for_attachment" method still computes the default method, forgetting that the fs_filename is the readable name and not obfuscated.
        
        I had to add the /dir1/dir2 to generate the correct URL

Screenshot from 2024-10-10 20-42-24
Screenshot from 2024-10-10 20-42-09
Screenshot from 2024-10-10 20-41-51

The module still needs testing and debugging, many tests failed and I see that store_fname is neglecting the directory path although it can be added like in my case.

I am looking forward to continuing debugging with assistance from the contributors.

Thanks All...

### DIFF
--- a/fs_attachment/tests/test_fs_storage.py
+++ b/fs_attachment/tests/test_fs_storage.py
@@ -1,5 +1,6 @@
 # Copyright 2023 ACSONE SA/NV (http://acsone.eu).
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+import base64
 import os
 
 from odoo.exceptions import ValidationError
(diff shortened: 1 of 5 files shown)


kobros-tech:
@OCA-git-bot 

@sebastienbeau 

@lmignon 

@mymage

fs_attachment/tests/test_fs_storage.py
```diff
@@ -372,3 +373,42 @@ def test_recompute_urls(self):
...
+    def test_url_for_image_dir_optimized_and_not_obfuscated(self):
+        # Create a base64 encoded mock image (1x1 pixel transparent PNG)
+        image_data = base64.b64encode(
+            b"\x89PNG\r\n\x1a\n\x00\x00\x00\rIHDR\x00\x00\x00\x01\x00\x00\x00\x01\x08"
+            b"\x06\x00\x00\x00\x1f\x15\xc4\x89\x00\x00\x00\nIDAT\x08\xd7c\xf8\x0f\x00"
+            b"\x01\x01\x01\x00\xd1\x8d\xcd\xbf\x00\x00\x00\x00IEND\xaeB`\x82"
+        )
+
+        # Create a mock image filestore
+        fs_storage = self.env["fs.storage"].create(
```
lmignon, line 386: What's the protocol of your storage? Protocol is a required field and there is nor more default value on it on the 16.0 branch. Be careful to don't serve your images from the odoo's filestore in a public way. If you configure your proxy to get access to the filestore under the 'images' path, you expose all the files from your odoo instance....

kobros-tech: @lmignon 

Thanks for clarifying, but I use different filestore for only products imsges thumbnails.

But the bug is completely different from these settings that you mentioned.

I can use many protocols and to test I used file protocol.

Thanks

lmignon (approved):
Thank you for the fix @kobros-tech LGTM (Code review only)

PR #408, opened 2024-10-17
[17.0][IMP] fs_product_multi_image: Replace tree view with kanban view

JasminSForgeFlow:


GuillemCForgeFlow (approved):
LGTM 👍🏿

fs_product_multi_image/views/product_template.xml
```diff
@@ -28,17 +28,25 @@
                                 nolabel="1"
                                 colspan="2"
                             >
-                                <tree>
-                                    <field name="sequence" widget="handle" />
-                                    <field name="name" />
-                                </tree>
+                            <kanban>
+                                <field name="name" />
```
lmignon, line 32: How do you manage sequence now? Do we really need to always display the images?  If you take a look at the code in version 16, a new widget has been developped with a lot of usefull functionnality. We should forward port https://github.com/OCA/storage/pull/332 instead.

lmignon (changes requested):
Thak you for the contribution. IMO we should forward port https://github.com/OCA/storage/pull/332

fs_product_multi_image/views/product_template.xml
JasminSForgeFlow: You are right

PR #412, opened 2024-10-25
[FIX] fs_attachment: remove unwanted args

kobros-tech:
The module works fine if the Point of Sale module is not installed.
Once I install the POS module, the server returns a 500 error. 
By debugging the error, I found that an argument called **"content_security_policy"** was passed into the sendFile function, and it had a value of **None** as this argument is passed without being managed.

As long as this argument and other unmanaged arguments do NOT affect the behavior of the module or even the system, they have to be removed.

Only specified arguments are fine to pass to the sendFile function.

There are two references to using the patched sndFile function, but only one prototype is passing unmanaged arguments and the original behavior here imitated the prototype behavior which caused the error.

I have documented the error, the solution, and the two types of good behavior in the Odoo code.

Screenshot from 2024-10-12 03-14-47

Screenshot from 2024-10-12 03-20-33

Screenshot from 2024-10-12 03-21-27

Screenshot from 2024-10-25 23-45-46

Screenshot from 2024-10-25 23-46-09

### DIFF
--- a/fs_attachment/README.rst
+++ b/fs_attachment/README.rst
@@ -423,6 +423,7 @@ Laurent Mignon <laurent.mignon@acsone.eu>
 Marie Lejeune <marie.lejeune@acsone.eu>
 Wolfgang Pichler <wpichler@callino.at>
 Nans Lefebvre <len@lambdao.dev>
+Mohamed Alkobrosli <alkobroslymohamed@gmail.com>
 
 Maintainers
 ~~~~~~~~~~~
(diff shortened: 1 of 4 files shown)


kobros-tech:
@lmignon 
Could you check this fix, please?

lmignon (approved):

lmignon:
ping @qgroulard

kafai-lam:
@kobros-tech would you like to check out #413 ?

Issue #411, opened 2024-10-24
[16.0] fs_attachment: Stored Compute Attachments Not Fetched from File Storage


miikanissi:
Due to the following condition, any binary field which is a stored compute or stored related field are not correctly fetched from the file storage:

https://github.com/OCA/storage/blob/eeeea37cb00b8205ee5db4db7f8b00645a3ba53c/fs_attachment/models/ir_binary.py#L28

Is there a reason why compute and related fields are ignored? Shouldn't it only care whether the file is stored in the database or not? 

This leads to an error if you have an attachment which is a stored compute field.

PR #414, opened 2024-11-05
[17.0][FIX] allow serving field attachment with fs stream

kafai-lam:
Current Issue: After the force_storage step is complete and the default file storage (fs.storage) is set, images are automatically migrated to use fs.attachment. However, image model fields such as image_128 and image_512 are not being served correctly by `ir.binary`. This PR aims to address this problem.

Proposed Solution: In the `IrBinary._get_image_stream_from` method, it is recommended to add a field attachment handler for binary fields when they are found to use fs.attachment. This should ensure that image fields are served properly.

### DIFF
--- a/fs_attachment/models/ir_binary.py
+++ b/fs_attachment/models/ir_binary.py
@@ -16,16 +16,14 @@ _logger = logging.getLogger(__name__)
 class IrBinary(models.AbstractModel):
     _inherit = "ir.binary"
 
-    def _record_to_stream(self, record, field_name):
-        # Extend base implementation to support attachment stored into a
-        # filesystem storage
-        fs_attachment = None
+    def _get_fs_attachment_for_field(self, record, field_name):
         if record._name == "ir.attachment" and record.fs_filename:
-            fs_attachment = record
+            return record
+
         record.check_field_access_rights("read", [field_name])
         field_def = record._fields[field_name]
-        if field_def.attachment and not field_def.compute and not field_def.related:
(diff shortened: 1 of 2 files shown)


fs_attachment/models/ir_binary.py
```diff
@@ -16,16 +16,14 @@
...
+    def _get_fs_attachment_for_field(self, record, field_name):
         if record._name == "ir.attachment" and record.fs_filename:
-            fs_attachment = record
+            return record
+
         record.check_field_access_rights("read", [field_name])
         field_def = record._fields[field_name]
-        if field_def.attachment and not field_def.compute and not field_def.related:
-            field_attachment = (
+        if field_def.attachment and field_def.store:
```
kafai-lam, line 25: I've changed the condition from "is not computed and not related" to "is stored". For instance, all image_64 and image_128 fields, defined by the image.mixin model, are computed fields but still have an actual record in ir.attachment.

This update may also help address the question raised in #411.

lmignon (approved):
@kafai-lam Thank you for the fix. The code LGTM (code review only)

lmignon:
@kafai-lam Can you rebase and fix the conflict plz. Once done, I'll proceed to the merge.

kafai-lam:
@lmignon Resolved conflicts in the test cases with rebase

lmignon:
> @lmignon Resolved conflicts in the test cases with rebase

Thank you.

PR #425, opened 2024-11-11
[17.0][FIX] fs_attachment: fix fs_url computing

kafai-lam:
Forward port changes from PR #404 to 17.0

CC @kobros-tech


lmignon:
@kafai-lam Thank you for the forward port. I did it with the oca-port command and this change is already in the [17.0](https://github.com/OCA/storage/blob/17.0/fs_attachment/models/fs_storage.py#L443) and [18.0](https://github.com/OCA/storage/blob/18.0/fs_attachment/models/fs_storage.py#L443) branches.

PR #428, opened 2024-11-13
[16.0][ADD] allow to download an image

mathieudelva:


lmignon:
@mathieudelva A more generic solution is provided by #429 . What do you think about?

PR #429, opened 2024-11-13
[16.0][IMP] fs_image: Add download button

lmignon:
On image hover, a download button is now available to easily download the image

image


bealdav (approved):
Just tested, it works like a charm.
No code review
Thanks a lot

PR #434, opened 2024-11-29
[18.0][MIG] storage_file: Migration to 18.0

thienvh332:


storage_file/models/storage_file.py
```diff
@@ -204,11 +200,11 @@ def unlink(self):
     def _clean_storage_file(self):
         # we must be sure that all the changes are into the DB since
         # we by pass the ORM
-        self.flush()
+        self.env.flush_all()
         self._cr.execute(
-            """SELECT id
+            SQL("""SELECT id
```
StefanRijnhart, line 205: Why is SQL applied? Is passing raw strings deprecated?

StefanRijnhart (commented):
Thanks!

StefanRijnhart:
Maybe squash 'Raph's commits into one big `[WIP] storage_file` commit
image

storage_file/models/storage_file.py
thienvh332: Just a solution to avoid SQL injections supported by odoo. You can still continue to use query by string.

storage_file/models/storage_backend.py
```diff
@@ -0,0 +1,171 @@
...
+        return "/".join([x.rstrip("/") for x in parts if x])
+
+    def _register_hook(self):
+        super()._register_hook()
+        backends = self.search([]).filtered(
+            lambda x: x._get_base_url_for_files() != x.base_url_for_files
+        )
+        if not backends:
+            return
+        sql = SQL(f"SELECT id FROM {self._table} WHERE ID IN %s FOR UPDATE")
```
StefanRijnhart: If you are applying `SQL` here because it protects better against SQL injection (see comment below), then you need to pass the table name as an `SQL.identifier`.

storage_file/models/storage_file.py
StefanRijnhart: OK, so it's your suggestion to apply `SQL` by default even if no arguments are passed here that could be used for sql injection. Fair enough.

StefanRijnhart (approved):
Thanks for the updates!

storage_file/models/storage_file.py
```diff
@@ -0,0 +1,230 @@
...
+        )
+    ]
+
+    def write(self, vals):
+        if "data" in vals:
+            for record in self:
+                if record.data:
+                    raise UserError(
+                        self.env._(
+                            "File can not be updated," "remove it and create a new one"
```
ivantodorovich:
```suggestion
                            "File can not be updated, remove it and create a new one"
```

storage_file/models/storage_file.py
```diff
@@ -0,0 +1,230 @@
...
+            super().unlink()
+        else:
+            self.write({"to_delete": True, "active": False})
+        return True
+
+    @api.model
+    def _clean_storage_file(self):
+        # we must be sure that all the changes are into the DB since
+        # we by pass the ORM
+        self.env.flush_all()
```
ivantodorovich:
```suggestion
        self.flush_model(["to_delete"])
```
nitpicking: this is actually more accurate

storage_file/tests/test_storage_file.py
```diff
@@ -232,12 +232,12 @@ def test_public_access1(self):
         # Public user used on the controller when authentication is 'public'
         public_user = self.env.ref("base.public_user")
         with self.assertRaises(AccessError):
-            # BUG OR NOT with_user doesn't invalidate the cache...
-            # force cache invalidation
-            self.env.cache.invalidate()
-            self.env[storage_file._name].with_user(public_user).browse(
-                storage_file.ids
-            ).name
+            _ = (
```
ivantodorovich: Why are you assigning to `_` ?

storage_file/models/storage_backend.py
```diff
@@ -0,0 +1,175 @@
...
+        return "/".join([x.rstrip("/") for x in parts if x])
+
+    def _register_hook(self):
+        super()._register_hook()
+        backends = self.search([]).filtered(
+            lambda x: x._get_base_url_for_files() != x.base_url_for_files
+        )
+        if not backends:
+            return
+        sql = SQL("SELECT id FROM %s WHERE ID IN %s FOR UPDATE")
```
ivantodorovich:
```suggestion
        sql = SQL(
            "SELECT id FROM %s WHERE ID IN %s FOR UPDATE",
            SQL.identifier(self._table),
            tuple(backends.ids),
        )
```

💡 tip: The nice thing about the `SQL` class (and probably why Odoo's introduced it) is that you can pass the parameters directly to each SQL instance, and they will later be combined together if you where to concatenate multiple SQL instances

not blocking of course, it's entirely optional. I'm just sharing this.

ivantodorovich (approved):

storage_file/models/storage_file.py
```diff
@@ -0,0 +1,230 @@
...
+            WHERE to_delete=True FOR UPDATE""")
+        )
+        ids = [x[0] for x in self._cr.fetchall()]
+        for st_file in self.browse(ids):
+            st_file.backend_id.sudo().delete(st_file.relative_path)
+            st_file.with_context(cleanning_storage_file=True).unlink()
+            # commit is required since the backend could be an external system
+            # therefore, if the record is deleted on the external system
+            # we must be sure that the record is also deleted into Odoo
+            st_file._cr.commit()
```
ivantodorovich, line 217: 💡 tip: This cron could be refactored to use the new `ir.cron._notify_progress` method

storage_file/tests/test_storage_file.py
thienvh332: This is why I declared the variable unused for it: `B018 Found useless attribute access. Either assign it to a variable or remove it.`

storage_file/models/storage_file.py
thienvh332: Thanks for sharing. I updated it :pray:

storage_file/tests/test_storage_file.py
ivantodorovich: Oh, if it's just a linter rule, you can disable it it with an inline comment.
E.g.:

```py
self.env[storage_file._name].with_user(public_user).browse(storage_file.ids).name  # noqa: B018
```

storage_file/models/storage_file.py
ivantodorovich: Actually the objective of using it would be to report progress whilst you process each item/batch of items.
This lets the framework to automatically re-scheduled the cron if it hasn't finished

See for example: https://github.com/odoo/odoo/blob/783aa718cab82f1289484f3187ad911c86c59358/addons/l10n_my_edi/models/account_move.py#L487
or https://github.com/odoo/enterprise/blob/31dbf40c85f1d6fb02bcb48d53e2a3f514e2e356/partner_commission/models/purchase_order.py#L83

It also lets you play with a LIMIT, to ensure each cron execution processes no more than X records.
It'd probably be interesting to do it here, too. In which case you'd need to identify the total number of records to process (with a search_count, for example); and then select for update only a batch of them (with a limit).

In any case. Not blocking over this.

ivantodorovich (approved):

storage_file/tests/test_storage_file.py
thienvh332: Thanks you, It doesn't work because it crossed the line when combined with pre-commit with max characters being 88 per line

lmignon: @thienvh332 Can you at least assign the value to a string starting with `_` to avoid the warning but not to the character `_` alone to avoid the confusion with the translation method `_`.

lmignon (dismissed):
Thank you for the migration. A little comment....

simahawk (approved):

PR #435, opened 2024-11-29
[18.0][MIG] storage_thumbnail: Migration to 18.0

thienvh332:
### Depends on:
- https://github.com/OCA/storage/pull/434


storage_thumbnail/models/thumbnail_mixin.py
```diff
@@ -0,0 +1,132 @@
...
+
+    def _get_medium_thumbnail(self):
+        return self.get_or_create_thumbnail(*self._image_scale_mapping["medium"])
+
+    def _get_small_thumbnail(self):
+        return self.get_or_create_thumbnail(*self._image_scale_mapping["small"])
+
+    def _get_url_key(self, url_key):
+        if url_key:
+            url_key = slugify(url_key)
```
ivantodorovich:
```suggestion
            url_key = self.env['ir.http']._slugify_one(url_key)
```

I wonder if we can use this method, instead, and drop the slugify dependency.
The core method uses the library when available, and fallsback to an internal implementation

storage_thumbnail/models/thumbnail_mixin.py
```diff
@@ -0,0 +1,132 @@
...
+        return True
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        records = super().create(vals_list)
+
+        for record in records:
+            record.generate_odoo_thumbnail()
+
+        return records
```
ivantodorovich:
```suggestion
        records = super().create(vals_list)
        for record in records:
            record.generate_odoo_thumbnail()
        return records
```

nitpicking: extra spaces

ivantodorovich (approved):

ajaniszewska-dev (approved):

PR #436, opened 2024-11-29
[18.0][MIG] storage_image: Migration to 18.0

thienvh332:
### Depends on:
- https://github.com/OCA/storage/pull/434
- https://github.com/OCA/storage/pull/435

PR #437, opened 2024-11-29
[18.0][MIG] storage_image_product: Migration to 18.0

thienvh332:
### Depends on:
- https://github.com/OCA/storage/pull/434
- https://github.com/OCA/storage/pull/435
- https://github.com/OCA/storage/pull/436


storage_image_product/models/image_tag.py
```diff
@@ -31,13 +31,11 @@ def _get_default_apply_on(self):
         comodel_name="product.image.relation",
         inverse_name="tag_id",
         string="Product Image Relations",
-        readonly=True,
```
ivantodorovich: Why are we removing these readonly attributes?
(same for `categ_img_rel_ids`)

thienvh332: Reason: `If a field is marked as readonly=True on the model, it won't be possible to import it through the Odoo import tool, so avoid it as possible and define the readonly attribute on the views instead.`
Document: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0

ivantodorovich (approved):

henrybackman (approved):

simahawk:
https://github.com/OCA/storage/pull/463

PR #440, opened 2025-01-06
[18.0][MIG] storage_backend_s3: Migration to 18.0

thienvh332:


storage_backend_s3/components/s3_adapter.py
```diff
@@ -0,0 +1,129 @@
...
+from odoo.addons.component.core import Component
+
+_logger = logging.getLogger(__name__)
+
+try:
+    import boto3
+    from botocore.exceptions import ClientError, EndpointConnectionError
+
+except ImportError as err:  # pragma: no cover
+    _logger.debug(err)
```
ivantodorovich, line 21: Since it's already declared in the manifest's `external_dependencies`, we can drop the try except block here

checklog-odoo.cfg
```diff
@@ -0,0 +1,3 @@
+[checklog-odoo]
+ignore=
+    WARNING odoo py.warnings.*ssl\.PROTOCOL_TLS\b is deprecated
```
ivantodorovich: What's this for? 🤔

ivantodorovich (approved):

checklog-odoo.cfg
thienvh332: A `warning` comes from the python library. I want to ignore it to avoid getting test errors from `oca-cli`.

ivantodorovich: Is it possible to upgrade the code, to align to the library recommendations?
If so that'd be preferable. Otherwise lets keep it like this

thienvh332: The problem comes from the library internals, not odoo. And changing the library version is also quite risky. So I ignore it.

ivantodorovich: Oh, I see this is pinned because of this issue:
- https://github.com/OCA/storage/issues/67 (https://github.com/OCA/storage/pull/218)

It's probably not a good idea to pin it to an outdated version forever, though. And the dependency problem should go away at some point (or at least let us bump a few versions).

Have you checked if the `urllib` common dependency issue still exists with Odoo 18.0?
At first glance, it seems Odoo now requires a higher version of urllib, there may not even be an issue

https://github.com/odoo/odoo/blob/18.0/requirements.txt#L79-L80

thienvh332: You're right, I updated it. Thanks :pray:

ivantodorovich: Thanks !

ajaniszewska-dev (approved):
seems good

lmignon (approved):

simahawk:
https://github.com/OCA/storage/pull/466

PR #443, opened 2025-02-05
[16.0][FIX] fs_attachment: Implement retry mechanism for file read failures:

AaronHForgeFlow:
Added a retry mechanism to `_storage_file_read` to handle intermittent `FileNotFoundError` issues. In real world this may happen when sending emails with heavy attachments on it. The file is attempted to be read but it is still not showing in S3. Tested with a mock file system that randomly raises `FileNotFoundError` to simulate real-world failures.

### DIFF
--- a/fs_attachment/models/ir_attachment.py
+++ b/fs_attachment/models/ir_attachment.py
@@ -373,14 +373,29 @@ class IrAttachment(models.Model):
     @api.model
     def _storage_file_read(self, fname: str) -> bytes | None:
         """Read the file from the filesystem storage"""
-        fs, _storage, fname = self._fs_parse_store_fname(fname)
-        try:
-            with fs.open(fname, "rb") as f:
-                return f.read()
-        except IOError:
-            _logger.info(
-                "Error reading %s on storage %s", fname, _storage, exc_info=True
-            )
+        fs, storage_code, fname = self._fs_parse_store_fname(fname)
+
+        read_retry_attempts, read_retry_delay = (
+            self.env["fs.storage"].sudo().get_read_retry_config(storage_code)
+            if storage_code
+            else (0, 0.0)
+        )
+        for attempt in range(read_retry_attempts + 1):
+            try:
+                with fs.open(fname, "rb") as f:
+                    return f.read()
+            except FileNotFoundError:
+                _logger.info(
+                    "File not found %s on storage %s (attempt %d)",
+                    fname,
+                    storage_code,
+                    attempt + 1,
+                )
+                time.sleep(read_retry_delay)
+            except IOError:
+                _logger.info(
+                    "Error reading %s on storage %s", fname, storage_code, exc_info=True
+                )
         return b""
 
     def _storage_write_option(self, fs):
--- a/fs_attachment/models/fs_storage.py
+++ b/fs_attachment/models/fs_storage.py
@@ -108,6 +108,16 @@ class FsStorage(models.Model):
         compute="_compute_field_ids",
         inverse="_inverse_field_ids",
     )
+    read_retry_attempts = fields.Integer(
+        default=0,
+        help="Number of times to retry reading a file before failing. "
+        "It can apply to certain file storage backends like S3.",
+    )
+    read_retry_delay = fields.Float(
+        string="Retry Delay (seconds)",
+        default=0.0,
+        help="Seconds to wait between retries when reading a file.",
+    )
 
     @api.constrains("use_as_default_for_attachments")
     def _check_use_as_default_for_attachments(self):
@@ -501,3 +511,13 @@ class FsStorage(models.Model):
         attachments = self.env["ir.attachment"].search(domain)
         attachments._compute_fs_url()
         attachments._compute_fs_url_path()
+
+    @api.model
+    @tools.ormcache("code")
+    def get_read_retry_config(self, code):
+        """Return (retry_attempts, retry_delay) for a given storage code."""
+        storage = self.get_by_code(code)
+        return (
+            max(storage.read_retry_attempts, 0) if storage else 0,
+            max(storage.read_retry_delay, 0.0) if storage else 0.0,
+        )
--- a/fs_attachment/views/fs_storage.xml
+++ b/fs_attachment/views/fs_storage.xml
@@ -36,6 +36,9 @@
                 <field name="is_directory_path_in_url" />
                 <field name="use_filename_obfuscation" />
                 <field name="use_x_sendfile_to_serve_internal_url" />
+                <separator string="Read Failure Parameters" />
+                <field name="read_retry_attempts" />
+                <field name="read_retry_delay" />
             </field>
         </field>
     </record>


lmignon (approved):

fs_attachment/models/ir_attachment.py
```diff
@@ -374,13 +376,22 @@ def _set_attachment_data(self, asbytes) -> None:  # pylint: disable=missing-retu
...
-        except IOError:
-            _logger.info(
-                "Error reading %s on storage %s", fname, _storage, exc_info=True
-            )
+        for attempt in range(FILE_READ_RETRIES):
+            try:
+                with fs.open(fname, "rb") as f:
+                    return f.read()
+            except FileNotFoundError:
+                _logger.info(
```
lmignon: The number of attempt and the delay could be properties of the `fs.storage` record.  It could make no sense to have a retry mechanism on others file systems than s3. Zero as default value for the new field used to control the number of attempt could disable the retry mechanism.

lmignon (changes requested):

fs_attachment/models/ir_attachment.py
AaronHForgeFlow: Yes, I totally agree so I did the change.

fs_attachment/models/ir_attachment.py
```diff
@@ -374,13 +374,28 @@ def _set_attachment_data(self, asbytes) -> None:  # pylint: disable=missing-retu
...
         """Read the file from the filesystem storage"""
         fs, _storage, fname = self._fs_parse_store_fname(fname)
-        try:
-            with fs.open(fname, "rb") as f:
-                return f.read()
-        except IOError:
-            _logger.info(
-                "Error reading %s on storage %s", fname, _storage, exc_info=True
-            )
+        read_retry_attempts = (
```
lmignon: ` _storage_file_read` is an `api.model` method. You should not access the storage via the `fs_strorage_id` field.
To get an instance of the `fs.storage` record you can use: 
```python
fs, storage_code, fname = self._fs_parse_store_fname(fname)
if storage_code:
    storage = self.env["fs.storage"].get_by_code(storage_code)

```

To avoid to always get access to the `fs.storage` record from the database, you could use a dedicated method decorated with `orm.cache`in the FSStorage class

```python
@api.model
@orm.cache("code")
def get_read_retry_config(self, code) -> tuple[float, float]:
    """ ...
    """
    storage = self.get_by_code(code)
    return max(storage.read_retry_attempts, 0), max(storage.read_retry_delay, 0.0)

**Review by lmignon — CHANGES_REQUESTED** [review]

File: fs_attachment/models/ir_attachment.py (hunk 2 above)
Comment (AaronHForgeFlow) [reply fs_attachment/models/ir_attachment.py]: Thank you for help @lmignon I adapted the code to not rely on self inside api.model method. I did some testing and it worked, it only access get_read_retry_config on the first attempt

File: fs_attachment/models/ir_attachment.py [hunk 3]
```diff
@@ -374,13 +374,28 @@ def _set_attachment_data(self, asbytes) -> None:  # pylint: disable=missing-retu
...
-        try:
-            with fs.open(fname, "rb") as f:
-                return f.read()
-        except IOError:
-            _logger.info(
-                "Error reading %s on storage %s", fname, _storage, exc_info=True
-            )
+
+        read_retry_attempts, read_retry_delay = (
+            self.env["fs.storage"].get_read_retry_config(_storage)
```
lmignon: Could you rename `_storage` to  `storage_code` the prefix `_` was used to mark the variable as unused into the previous implementation. It's no more the case and it's real meaning is `storage_code`.

Thank you for all your work.

AaronHForgeFlow: Sure. Thank you for pointing that out. It is now renamed.

lmignon (approved):

lmignon:
@sbidoul @adrienpeiffer @qgroulard

qgroulard (approved):
LGTM (code review)

fs_attachment/models/ir_attachment.py
```diff
@@ -373,14 +373,29 @@ def _set_attachment_data(self, asbytes) -> None:  # pylint: disable=missing-retu
...
+                with fs.open(fname, "rb") as f:
+                    return f.read()
+            except FileNotFoundError:
+                _logger.info(
+                    "File not found %s on storage %s (attempt %d)",
+                    fname,
+                    storage_code,
+                    attempt + 1,
+                )
+                time.sleep(read_retry_delay)
```
sbidoul, line 394: So this means FileNotFoundError will never be raised anymore ?

fs_attachment/models/ir_attachment.py
```diff
@@ -373,14 +373,29 @@ def _set_attachment_data(self, asbytes) -> None:  # pylint: disable=missing-retu
...
+            except IOError:
+                _logger.info(
+                    "Error reading %s on storage %s", fname, storage_code, exc_info=True
+                )
```
sbidoul, line 398: And in case of IOError we return an empty byte string (even before this PR). Is that intended @lmignon ?

sbidoul:
Also, regarding the implementation, I'd suggest doing the retry mechanism in a context manager that can be unit tested.

```python
with retrying():
   ...
```
 
I can recommend the [tenacity](https://tenacity.readthedocs.io), although we may not want to pull and additional dependency just for that.

sbidoul (changes requested):
I have questions :)

fs_attachment/models/ir_attachment.py
AaronHForgeFlow: For my customer this happens for example when sending emails from invoices with attachments on it, and because of this error the pdf is empty in the email that is sent to the customer. But if a second email is send with the same attachment it succeeds. 

I suspect the file was just uploaded and S3 did not have the time to list the file in all the nodes. That explains why the attachment is accessible a few moments later. I will do more tests today on real env.

AaronHForgeFlow:
> Also, regarding the implementation, I'd suggest doing the retry mechanism in a context manager that can be unit tested.
> 
> ```python
> with retrying():
>    ...
> ```
> 
> I can recommend the [tenacity](https://tenacity.readthedocs.io), although we may not want to pull and additional dependency just for that.

I will look into it. Thank you @sbidoul

fs_attachment/models/ir_attachment.py
lmignon: Yes it's indended. When trying to get access to files stored into an external filesystem, we don't want to crash if the files cannot be accessed. This is necessary to avoid blocking Odoo's operation if an external system is not accessible. And we mimic what odoo does natively https://github.com/odoo/odoo/blob/18.0/odoo/addons/base/models/ir_attachment.py#L124

sbidoul: I see, thanks. Surprising.

fs_attachment/models/ir_attachment.py
sbidoul: > So this means FileNotFoundError will never be raised anymore ?

Ok, so since FileNotFoundError is a subclass of IOError, it will be the same as before and return `b""`.

sbidoul:
So my questions are answered, thanks.

Still it's a bit surprising that this comes up only now. Which S3 provider do you use @AaronHForgeFlow ?

AaronHForgeFlow:
> So my questions are answered, thanks.
> 
> Still it's a bit surprising that this comes up only now. Which S3 provider do you use @AaronHForgeFlow ?

I am using AWS and yes, it is rare that this did not happen before. I hope this fixes the issue I am having. I will provide some feedback here soon.

AaronHForgeFlow:
After testing this in production I found that this PR fixed the issue only partially, there are some other filesystem.open calls in the same file that are raising errors when uploading invoices by drag & dropping on the invoices:

https://github.com/OCA/storage/blob/16.0/fs_attachment/models/ir_attachment.py#L1019

https://github.com/OCA/storage/blob/16.0/fs_attachment/models/ir_attachment.py#L1054

I implemented also the retry mechanism in those calls too, however I am not sure if there is another underlying issue. It is rare this is not happening in other instances. I am using protocol S3a(S3 filesystem), same as in other instances not having this issue. I am doing this in odoo.sh, but I don't think that that is relevant. Any comments will be appreciated.

AaronHForgeFlow:
Definitely, retrying is not the solution, as all the retries attempts fail. I am closing this PR.

I suspect that caching is interfering with the file lookups, I am looking into that now.

Thank you all for the help.

PR #445, opened 2025-02-19
[18.0][MIG] fs_file

rousseldenis:


rousseldenis:
@lmignon

lmignon (approved):
LGTM (Code review + diff with 16.0 and functional via #449)

PR #453, opened 2025-03-04
[18.0][FIX] fs_storage: fix `_ls_check_connection` when dealing with huge amount of files in root folder

sebalix:
### Context

We tried to configure a storage on an `AzureBlobFileSystem` with 3.8+M of files located at the root directory.

### Issue

This storage is in read-only access, so we have to use the other option `ls` to check the connection. Issue is that the current `ls` on root folder will try to return too much file names, increasing the consumed memory, leading to a kill of the Odoo process.

### Proposed solution

Perform this `ls` on a non-existing file, which should then raise `FileNotFoundError` (to check if this exception is returned by all protocols?). If nothing is raised, that also means the connection is working (as before).

EDIT: Feedback, it is working well after 2 weeks

### DIFF
--- a/fs_storage/models/fs_storage.py
+++ b/fs_storage/models/fs_storage.py
@@ -21,6 +21,8 @@ from odoo.addons.base_sparse_field.models.fields import Serialized
 
 _logger = logging.getLogger(__name__)
 
+LS_NON_EXISTING_FILE = ".NON_EXISTING_FILE"
+
 
 # TODO: useful for the whole OCA?
 def deprecated(reason):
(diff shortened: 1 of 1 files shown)


fs_storage/models/fs_storage.py
```diff
@@ -21,6 +21,8 @@
 
 _logger = logging.getLogger(__name__)
 
+LS_NON_EXISTING_FILE = ".NON_EXISTING_FILE"
```
sebalix, line 24: <s>Should it be a new config field? Should we add some random characters when using it to ensure it is non-existing on remote FS?</s>

sebalix: (even if the file exists, as soon as `ls` returns we are good)

henrybackman (approved):

simahawk (approved):
LG

lmignon (approved):

PR #455, opened 2025-03-10
[16.0][FIX] fs_storage: Fixed issue with fsspec version 2025.3.0

baguenth:
This pr fixes https://github.com/OCA/storage/issues/454

PR #457, opened 2025-03-11
[FIX] fs_storage: Add a KeyError Exception

etobella:
This is necessary in order to avoid issues in case fsspec added a new protocol and forget the error key

Without this change, edi-framework and storage are failing (and other tests might happen too).

The real error  is here:

https://github.com/fsspec/filesystem_spec/blame/50c5a2e161d85413e19e9e10d42de8b3cf56a921/fsspec/registry.py#L75-L77

An err key should be added, but it was forgotten. I created a PR in fsspec in order to fix it, but I think we can add this in order to avoid other errors.

https://github.com/fsspec/filesystem_spec/pull/1804

@lmignon @simahawk

### DIFF
--- a/fs_storage/models/fs_storage.py
+++ b/fs_storage/models/fs_storage.py
@@ -240,7 +240,7 @@ class FSStorage(models.Model):
             try:
                 cls = fsspec.get_filesystem_class(p)
                 protocol.append((p, f"{p} ({cls.__name__})"))
-            except ImportError as e:
+            except Exception as e:
                 _logger.debug("Cannot load the protocol %s. Reason: %s", p, e)
         return protocol
 
@@ -274,7 +274,7 @@ class FSStorage(models.Model):
             try:
                 fsspec.get_filesystem_class(p)
                 protocol.append((p, p))
-            except ImportError as e:
+            except Exception as e:
                 _logger.debug("Cannot load the protocol %s. Reason: %s", p, e)
         return protocol
 


kluna1998 (approved):
LGTM! 👍🏽

fs_storage/models/fs_storage.py
```diff
@@ -242,6 +242,8 @@ def _get_protocols(self) -> list[tuple[str, str]]:
                 protocol.append((p, f"{p} ({cls.__name__})"))
             except ImportError as e:
```
luisDIXMIT: Could this syntax be more efficient?
```suggestion
            except (ImportError, KeyError) as e:
                   _logger.debug("Cannot load the protocol %s. Reason: %s", p, e)
```

luisDIXMIT (approved):
LGTM!

lmignon (commented):
Do we need to circumvent a bug into the lib or can we just avoid to use the dependency's version of fsspec introducing the bug as proposed in https://github.com/OCA/storage/pull/455

etobella:
Well, I will use this PR in my internal implementations for sure with gitaggregate... once the library is fixed we will not need it anymore probably. Enforcing the library might work locally, but might give problems in other systems (it implies that everyone checks the dependencies on their projects and force the right one)

baguenth:
I raised https://github.com/OCA/storage/pull/455 but actually think, that the solution proposed here is more general.

baguenth (approved):

lmignon:
> I raised #455 but actually think, that the solution proposed here is more general.

If we want a more general solution we should catch  all 'Exception' instances to avoid to add another kind of exception in the future for not yet known error...

fs_storage/models/fs_storage.py
```diff
@@ -240,7 +240,7 @@ def _get_protocols(self) -> list[tuple[str, str]]:
             try:
                 cls = fsspec.get_filesystem_class(p)
                 protocol.append((p, f"{p} ({cls.__name__})"))
-            except ImportError as e:
+            except (ImportError, KeyError) as e:
```
lmignon:
```suggestion
            except Exception as e:
```

fs_storage/models/fs_storage.py
```diff
@@ -274,7 +274,7 @@ def _get_options_protocol(self) -> list[tuple[str, str]]:
             try:
                 fsspec.get_filesystem_class(p)
                 protocol.append((p, p))
-            except ImportError as e:
+            except (ImportError, KeyError) as e:
```
lmignon:
```suggestion
            except Exception as e:
```

lmignon (changes requested):
Catching the ImportError is not the right approach since it assume that the implementation will always raise this kind of exception when it's not possible to load a specific protocol. Any exception occurring at this stage has the same consequence.... the protocol can't be loaded

lmignon (approved):

PR #458, opened 2025-03-11
[17.0][FIX] fs_storage: Add a generic Exception

lmignon:
This is necessary in order to avoid issues in case fsspec added a new protocol and forget the error key

FWP #457


lmignon (commented):
/ocabot merge patch

PR #460, opened 2025-03-20
[18.0][IMP] fs_attachement, fs_storage: Move storage to model / field association to base addon

lmignon:


hparfr (approved):

hparfr:
What is the rational about this change ?

lmignon:
> What is the rational about this change ?

I'm developing a new module in which I need to know which storage should be used for a new field type.

sbidoul (approved):

PR #461, opened 2025-03-21
[18.0][FIX]: fs_storage fix _ls_check_connection enforce searching on dir

leemannd:
The goal of the fix done in -> https://github.com/OCA/storage/pull/453 is to search for a non existing file.

It is not compatible as is with s3fs which will search recursively in all folders.
I'm enforcing a dir at the root of the bucket to ensure the search will not be done recursively.

Instead of using `fs.ls(LS_NON_EXISTING_FILE, detail=False)` , I'm currently wondering if we couldn't use `fs.exists(LS_NON_EXISTING_FILE)`

### DIFF
--- a/fs_storage/models/fs_storage.py
+++ b/fs_storage/models/fs_storage.py
@@ -21,7 +21,7 @@ from odoo.addons.base_sparse_field.models.fields import Serialized
 
 _logger = logging.getLogger(__name__)
 
-LS_NON_EXISTING_FILE = ".NON_EXISTING_FILE"
+LS_NON_EXISTING_FILE = "/.NON_EXISTING_FILE/a"
 
 
 # TODO: useful for the whole OCA?


sebalix:
Good idea for `.exists()`, maybe there was a reason to use `ls` in the initial implementation.

leemannd:
Hello @florian-dacosta Do you have an opinion on it?

florian-dacosta:
Hello,
I did not really implement the `ls` implementation, but actually I did re-implement it because it had been removed.
So I did not the first implementation. I did not  check but it must be @lmignon, if you have an opinion on this one ?

On my side, I have no argument against the change.

florian-dacosta (changes requested):
Still, I did test this PR against 3 production used (s)ftp.

It did fail for one of them.
It does work with current version (where `LS_NON_EXISTING_FILE = ".NON_EXISTING_FILE"`)
The failing storage is a SFTP with `directory_path : "/"`
And I have the following error : `Path /.NON_EXISTING_FILE/a is not a subpath of the root path `

Also I wonder if the new `LS_NON_EXISTING_FILE` does work no matter what is set on the `directory_path` field (seems it does not at least in one of my use case.)

gurneyalex:
Maybe we need the check connection method to be overridable / overriden by backend?

lmignon:
> Still, I did test this PR against 3 production used (s)ftp.
> 
> It did fail for one of them. It does work with current version (where `LS_NON_EXISTING_FILE = ".NON_EXISTING_FILE"`) The failing storage is a SFTP with `directory_path : "/"` And I have the following error : `Path /.NON_EXISTING_FILE/a is not a subpath of the root path `
> 
> Also I wonder if the new `LS_NON_EXISTING_FILE` does work no matter what is set on the `directory_path` field (seems it does not at least in one of my use case.)

If the same method give different result depending of the protocol,  IMO it's a bug into the fsspec lib and should be addressed at this level.

PR #475, opened 2025-05-07
[18.0][MIG] Migration of fs_storage_backup

tarteo:
from 16.0


tonygalmiche (approved):
Tested on a local Odoo 18.0 instance with a local filesystem storage: backup generation (including filestore), automatic cleanup of old backups based on retention, and error handling (logged to chatter, no crash) all work as expected.

PR #478, opened 2025-06-02
[FIX] fs_storage: test connection by creating a hidden file

kobros-tech:
the lis opton in test connection is fine, but the create option was not functionnning correctly.

### DIFF
--- a/fs_storage/models/fs_storage.py
+++ b/fs_storage/models/fs_storage.py
@@ -290,7 +290,11 @@ class FSStorage(models.Model):
         return ".odoo_fs_storage_%s.marker" % self.id
 
     def _marker_file_check_connection(self, fs):
+        if not self.directory_path:
+            raise ValidationError(_("Please set a 'Directory Path' value"))
         marker_file_name = self._get_marker_file_name()
+        bucket = self.directory_path
+        fs.makedirs(bucket, exist_ok=True)
         try:
             fs.info(marker_file_name)
         except FileNotFoundError:
--- a/fs_storage/README.rst
+++ b/fs_storage/README.rst
@@ -268,6 +268,7 @@ Contributors
 
 * Laurent Mignon <laurent.mignon@acsone.eu>
 * Sébastien BEAU <sebastien.beau@akretion.com>
+* Mohamed Alkobrosli <malkobrosly@kencove.com>
 
 Maintainers
 ~~~~~~~~~~~
--- a/fs_storage/readme/CONTRIBUTORS.rst
+++ b/fs_storage/readme/CONTRIBUTORS.rst
@@ -1,2 +1,3 @@
 * Laurent Mignon <laurent.mignon@acsone.eu>
 * Sébastien BEAU <sebastien.beau@akretion.com>
+* Mohamed Alkobrosli <malkobrosly@kencove.com>
--- a/fs_storage/static/description/index.html
+++ b/fs_storage/static/description/index.html
@@ -512,7 +512,7 @@ main configuration file as follows:</p>
 <pre class="code ini literal-block">
 <span class="k">[fs_storage.fsprod]</span><span class="w">
 </span><span class="na">protocol</span><span class="o">=</span><span class="s">s3</span><span class="w">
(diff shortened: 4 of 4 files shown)


kobros-tech:
https://github.com/user-attachments/assets/96535d1a-1264-4983-9551-34f930286e7b

kobros-tech:
@mymage
@rvalyi 
@davidbeckercbl
@etobella 
@lmignon 
@sebastienbeau

kobros-tech:
@wlin-kencove

etobella:
This is only failing on s3. I am not sure if it makes sense (also, you are requiring a test bucket, I don't know if it makes sense).

kobros-tech:
> bucket

ok, I am using s3 cloudflare in my test a real prod. s3
my employer uses aws s3

I agree with no it does not make sense, but why do't we approve the commit?

fs_storage/models/fs_storage.py
```diff
@@ -290,11 +290,14 @@ def _get_marker_file_name(self):
         return ".odoo_fs_storage_%s.marker" % self.id
 
     def _marker_file_check_connection(self, fs):
+        bucket_name = "test"
+        fs.makedirs(bucket_name, exist_ok=True)
```
lmignon: Sorry but I don't see the point here. This code makes no sens to me.

lmignon (changes requested):

etobella (changes requested):
I think it is not necessary

fs_storage/models/fs_storage.py
kobros-tech: what if someone run test connection and get the error?

kobros-tech: the point is you have to specify dir name not just file name

if you don't like creating a test dir, then there must be a value for path directory and be used in test connection.

in any case dir name must be specified.

etobella: In this case, it is more useful to add a comment in order to ask the user to check both of them, as there are false negatives some times

kobros-tech: I will try something more meaningful

kobros-tech:
@etobella 
@lmignon 

I think this one is more meaningful than the previous approach?

fs_storage/models/fs_storage.py
```diff
@@ -290,7 +290,11 @@ def _get_marker_file_name(self):
         return ".odoo_fs_storage_%s.marker" % self.id
 
     def _marker_file_check_connection(self, fs):
+        if not self.directory_path:
+            raise ValidationError(_("Please set a 'Directory Path' value"))
```
lmignon, line 294: The directory_path is not required and should not be. In the case of a s3 storage, the directory_path is the bucket name  and it must exist.

lmignon (changes requested):
Sorry but, when using S3, your bucket must exist. It's not the responsibility of the `fs.storage` to create it. Moreover, the directory path must remain an optional parameter.

fs_storage/models/fs_storage.py
kobros-tech: ok, the backup module is creating backups dir, it can be the root "bucket" or can be child dir of directory path.

so fs.storage is creating a backet in this case!

but in case there is no bucket created yet we would create or notify to create there?

In my opinion it is tough to ask a user to ceate a bucket, but it would be the option that satisfys you.

if yes, then I would select the first bucket from the list that I may find and use it to ceate the test file, what do you think?

lmignon: Does your backup module create and configure the storages? If it's the case, it should be the responsibility of your addon to create the directory. But may be I don't understand the problem you are facing with the existing code. When you specify a directory path into the storage definition it must exist. The root directory could be a sensitive information and requires some specific privilege to create it. On a security point of view, the Odoo process should not have access to the parent folder of the specified directory path. That's why the auto creation seems to me a bad idea and must be avoided.

kobros-tech: I respect your point of view, the reason why I got my idea unintentionally is this line:

https://github.com/OCA/storage/blob/16.0/fs_storage_backup/models/fs_storage.py#L66

as I saw the backup module creating the dir, I decided to follow.

In order to finish this with a progress, what do you suggest to avoid raising that error, or just ignore and close this PR?

lmignon: @kobros-tech  IMO we should close this PR

PR #480, opened 2025-06-03
[18.0][ADD] fs_folder: An new addon to link a folder accessed through the fsspecs library to an odoo record.

lmignon:
If you need to link some specific models to a specific folder into an external filesystem and be able to manage the content of this folder from the model form view, the `fs_folder` addon is for you.
image

With the contribution of @etobella


etobella:
@lmignon on the rename, all commit history will be lost on next version. There are options to rename the history or you can just squash all commits together and add coauthors in order to simplify everything (second option is probably the easier one)

lmignon:
> @lmignon on the rename, all commit history will be lost on next version. There are options to rename the history or you can just squash all commits together and add coauthors in order to simplify everything (second option is probably the easier one)

No need to do it :smirk:; the path into the history is also renamed...
image

etobella (approved):

youring:
Thank you for creating this very useful module! It seems very similar to Alfodoo project. Will version control be added in the future?

Tested on runboat, found two points: 
If create a folder for the first time, 
`ValueError: No default storage found for the content of the external filesystem fields for model res.partner and field fs_folder_field. Please set a default storage in the filesystem storage configuration.`
Go to **FS Storage**, set **Odoo Filesystem Backend** as **Default For Fs Contents**, now it's fixed.
Should this be automatically set by demo module?

It seems preview is handled by the standalone SimpleDialog method, image is not centered, with white frame, and it can't be closed with the ESC key.
preview_img_not_centered_white_bg
And it's unable to navigate multiple files in current folder using the left and right arrow keys like Odoo core's preview.
A dumb question: for consistency, can we leverage Odoo core's preview component for attachment preview to replace SimpleDialog here?

lmignon:
> Thank you for creating this very useful module! It seems very similar to Alfodoo project. 

Same company, same authors ... 

> Will version control be added in the future?

The “fs_folder” module allows you to manage content stored in file systems external to Odoo from within Odoo. In its basic version, it focuses solely on cross-functional features for all file systems supported by the Python fsspecs library, which we use as an abstraction layer for the various file systems. Document versioning is not one of these standard features and is therefore not offered. However, the module has been designed to be easily extensible. An initial specialisation offered concerns more specific support for Microsoft Drive ( and SharePoint) directories. In the latter case, the target file system offers document version management features. These could be exposed in this specific context, even if it is not currently a requirement expressed by our customers. But do we really need it into Odoo? Indeed, It's very easy for users to jump from Odoo to a specific file or directory in the Microsoft application that manages these items and benefit from these advanced features without having to complicate the implementation in Odoo.

> 
> Tested on runboat, found two points: If create a folder for the first time, `ValueError: No default storage found for the content of the external filesystem fields for model res.partner and field fs_folder_field. Please set a default storage in the filesystem storage configuration.` Go to **FS Storage**, set **Odoo Filesystem Backend** as **Default For Fs Contents**, now it's fixed. Should this be automatically set by demo module?

It could

> 
> It seems preview is handled by the standalone SimpleDialog method, image is not centered, with white frame, and it can't be closed with the ESC key. preview_img_not_centered_white_bg And it's unable to navigate multiple files in current folder using the left and right arrow keys like Odoo core's preview. A dumb question: for consistency, can we leverage Odoo core's preview component for attachment preview to replace SimpleDialog here?

We have not found a simple way to use the native Odoo component. The preview could be improved, and all contributions are welcome. At this stage, we are not using the basic version of the `fs_folder` addon; we are using the `fs_folder_ms_drive` addon. Nevertheless, we have taken care to design this new document management approach to make it as modular as possible. As mentioned above, `fs_folder` is the basic component, and although it already offers many features, some of them, such as the preview, are offered in this first draft in a simple form because they are not used in our projects (in the specialization for ms_drive, we use the preview system provided by Microsoft).

PR #442, opened 2025-01-31
[17.0][IMP] fs_storage: Allow setting check_connection_method in configuration file

vincent-hatakeyama:

PR #485, opened 2025-06-13
[18.0][IMP] fs_storage: Allow setting check_connection_method in configuration file

vincent-hatakeyama:
Port of #442 to 18.0


etobella (commented):
Making this change will loose the data from the instance, as the system will drop the column check_connection_method.

At least we need a migration script. The same might happen with the original version.

vincent-hatakeyama:
> At least we need a migration script. The same might happen with the original version.

I’ve added an update script.

lmignon (approved):

etobella (approved):

PR #486, opened 2025-06-17
16.0 fs attachment metadata

ayushin:
This module update remote backend metadata with mimetype so that the files can later be served with signed URLS

### DIFF
--- a/fs_attachment_metadata/models/ir_attachment.py
+++ b/fs_attachment_metadata/models/ir_attachment.py
@@ -1,45 +1,51 @@
 # models/ir_attachment.py
 import logging
+import os
 
-from odoo import models
-from odoo.addons.fs_attachment.models.ir_attachment import AttachmentFileLikeAdapter
+from odoo import models, api
 
 _logger = logging.getLogger(__name__)
-
-
-class AttachmentFileLikeAdapterWithMetadata(AttachmentFileLikeAdapter):
-    """Enhanced adapter that passes metadata when opening files for writing."""
-
-    def _open_fs_file(self, mode):
-
-        print("OPENFILE")
-        """Override to pass metadata when opening files."""
-        kwargs = {}
-        if 'w' in mode and self.attachment.mimetype:
-            kwargs['metadata'] = {'contentType': self.attachment.mimetype}
-            _logger.debug(
-                "Opening file %s with content-type metadata: %s",
-                self.fs_filename,
-                self.attachment.mimetype
-            )
-
-        return self.fs.open(self.fs_filename, mode, **kwargs)
 
 
 class IrAttachment(models.Model):
     _inherit = "ir.attachment"
 
+    def _get_datas_related_values(self, data, mimetype):
+        """Override to pass mimetype to storage operations."""
+        # Store mimetype in context for _storage_file_write to access
+        if mimetype:
+            self = self.with_context(attachment_mimetype=mimetype)
+
+        # Call parent method which will eventually call _storage_file_write
+        return super()._get_datas_related_values(data, mimetype)
+
+    @api.model
+    def _storage_file_write(self, bin_data: bytes) -> str:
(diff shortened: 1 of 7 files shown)


fs_attachment_metadata/models/ir_attachment.py
```diff
@@ -0,0 +1,66 @@
...
+        # Store mimetype in context for _storage_file_write to access
+        if mimetype:
+            self = self.with_context(attachment_mimetype=mimetype)
+
+        # Call parent method which will eventually call _storage_file_write
+        return super()._get_datas_related_values(data, mimetype)
+
+    @api.model
+    def _storage_file_write(self, bin_data: bytes) -> str:
+        """Write the file to the filesystem storage with content-type metadata."""
```
lmignon, line 24: Your method must call super otherwise you breaks the inheritance chain. Moreover this method is almost a complete duplicate of the existing one.

lmignon (changes requested):
@ayushin Thank you for your proposal. 

As I don't really understand the problem to be solved or the context in which it occurs, it's difficult for me to understand the implementation choices that are made and to be able to make an in-depth review of your proposal. As it stands, however, there are a number of blocking elements, as pointed out in one of the comments. Don't hesitate to add context and use cases in the module description. This will help people to know whether it can be useful to them and will also facilitate the review.

fs_attachment_metadata/models/ir_attachment.py
ayushin: @lmignon i can create an issue. 

the point here is that if we want to serve the content of the file directly from an external url (and for example use pre-signed urls) and we have filename obfuscation we need to store the contet-type metadata on the external storage otherwise the datas won't be served correctly

the reason i don't call super is that i completely override the parent method as the limitation of the methods decomposition of the architecture does not allow more granular override

a better way would be to integrate metadata handling into the fs_file itself

lmignon: @ayushin Thank you for the explanation. I've to think about your issue. I've the feeling a better solution should be possible and more generic. It's a quite complex subject and i's in line with discussions we've been having internally to improve Odoo process load balancing, notably through the X-sendfile mechanism, but for protected resources. IMO, if your files are obfuscated, it's also important to not declare the mimetype to avoid to ease the access to it's content.

ayushin: well, our use case is to serve the content directly from the bucket via signed URLs rather than using x-sendfile that still proxies the binary data which is a huge performance penalty on large files (our case)

therefore the obfuscated files are not publicly readable - you need to sign the URL to access them (but we use obfuscation for reasons other than access)

you clearly shouldn't be relying on obfuscation for access

lmignon: The only difference I see in your overwrite is the injection of the mimetype  into kwargs from the context.....
But you can do the same by simply overriding the `_storage_write_option` method.  In it's original implementation it already do the same kind of process for s3 https://github.com/OCA/storage/blob/16.0/fs_attachment/models/ir_attachment.py#L386C1-L394C18

```python
    def _storage_write_option(self, fs):
        _fs = fs
        mimetype = self.env.context.get("mimetype")
        if mimetype:
            while _fs:
                if hasattr(_fs, "s3"):
                    return {"ContentType": mimetype}
                _fs = getattr(_fs, "fs", None)
        return {}
```

ayushin:
Thanks that's very helpful!

lmignon:
@ayushin Can we close this PR?

lmignon:
@ayushin Feel free to reopen-it if you still think it's useful.
Thank you for your proposal.

fs_attachment_metadata/models/ir_attachment.py
ayushin: Yes, that's the right way to do that, also for GCS

PR #497, opened 2025-07-24
[16.0] [Add] fs_attachment_s3: support X-Accel-Redirect for S3-backed attachments

lmignon:
Add fs_attachment_s3: support X-Accel-Redirect for authenticated S3 storage

This PR introduces a new addon, fs_attachment_s3, which simplifies the use of the X-Accel-Redirect mechanism to serve attachments stored on S3-compatible backends.

It enables X-Accel-Redirect even when attachments are stored on authenticated S3-like servers. This is achieved by having Odoo generate a signed internal URL that the reverse proxy (e.g., Nginx or Apache) can use to fetch and serve the file securely, without exposing credentials.

Example: Nginx configuration
If you're using Nginx as a reverse proxy and have enabled the "X-Sendfile to serve internal URL" option on your fs.storage configured for S3, you only need to add the following directive to your Nginx configuration. After that, Odoo will no longer serve the file content—Nginx will serve it directly:

```nginx
location ~ ^/fs_x_sendfile/(.*?)/(.*?)/(.*) {
    internal;

    set $url_scheme $1;
    set $url_host $2;
    set $url_path $3;
    set $url $url_scheme://$url_host/$url_path;

    proxy_pass $url$is_args$args;
    proxy_set_header Host $url_host;
    proxy_ssl_server_name on;
}
```

This improves performance by offloading file transfer to the reverse proxy, while maintaining secure access through signed URLs generated by Odoo in the case of private S3.


lmignon:
FYI @sbidoul 

I should maybe create a specific addon for the code specific to s3....

fs_attachment/models/ir_attachment.py
```diff
@@ -677,6 +678,13 @@ def _get_storage_codes(self):
         """Get the list of filesystem storage active in the system"""
         return self.env["fs.storage"].sudo().get_storage_codes()
 
+    def _get_x_accel_redirect_path(self):
+        """Get the path to use for X-Accel-Redirect"""
+        self.ensure_one()
+        if not self.fs_url_path:
+            raise UserError(_("The attachment is not stored in a filesystem storage."))
```
sbidoul: That's probably not a user mistake?

fs_attachment/models/fs_storage.py
```diff
@@ -501,3 +505,39 @@ def recompute_urls(self) -> None:
...
         attachments._compute_fs_url_path()
+
+    @api.model
+    def _get_x_accel_redirect_path(self, attachment: IrAttachment):
+        """Get the path to use for X-Accel-Redirect"""
+        if not attachment.fs_url_path:
+            raise RuntimeError(
+                "The attachment %s is not stored in a filesystem storage."
+                % attachment.id
+            )
```
sbidoul: This error should come from the super() call?

fs_attachment/models/fs_storage.py
```diff
@@ -501,3 +505,39 @@ def recompute_urls(self) -> None:
...
+                "get_object",
+                Params={"Bucket": fs_storage.directory_path, "Key": url_path},
+                ExpiresIn=3600,
+            )
+            # exclude the endpoint_url from the path
+            # for s3 storage, we use a signed URL
+            endpoint_url = s3_client.meta.endpoint_url
+            path /= url.replace(endpoint_url, "").lstrip("/")
+        else:
+            path /= attachment.fs_url_path.lstrip("/")
```
sbidoul: call super?

sbidoul (commented):
> I should maybe create a specific addon for the code specific to s3....

That's probably cleaner indeed.

fs_attachment/models/fs_storage.py
```diff
@@ -501,3 +505,39 @@ def recompute_urls(self) -> None:
...
+            url = self._s3_call_generate_presigned_url(
+                s3_client,
+                "get_object",
+                Params={"Bucket": fs_storage.directory_path, "Key": url_path},
+                ExpiresIn=3600,
+            )
+            # exclude the endpoint_url from the path
+            # for s3 storage, we use a signed URL
+            endpoint_url = s3_client.meta.endpoint_url
+            path /= url.replace(endpoint_url, "").lstrip("/")
```
sbidoul: We may need a specific prefix here so the nginx config can define an *internal* `location` that proxies the S3 endpoint.

lmignon: > We may need a specific prefix here so the nginx config can define an _internal_ `location` that proxies the S3 endpoint.

The prefix is the storage_code `path = Path("/") / attachment.fs_storage_code`

fs_attachment_s3/README.rst
```diff
@@ -0,0 +1,121 @@
...
+On the Odoo instance, go to *Settings* > *Technical* > *Storage* > *File
+Storage*.
+
+When you create a new storage for s3 or modify an existing one, when you
+activate the option "Use X-Sendfile To Serve Internal Url", 2 additional
+fields will appear:
+
+- **S3 Uses Signed URL For X-Accel-Redirect**: If checked, the
+  X-Accel-Redirect path will be a signed URL, which is useful for S3
+  storages that require signed URLs for access.
```
sbidoul, line 58: Worth explaining how the returned path is structured.

fs_attachment_s3/models/fs_storage.py
```diff
@@ -0,0 +1,82 @@
...
+            url = self._s3_call_generate_presigned_url(
+                s3_client,
+                "get_object",
+                Params={"Bucket": fs_storage.directory_path, "Key": file_path},
+                ExpiresIn=fs_storage.s3_signed_url_expiration,
+            )
+            # exclude the endpoint_url from the path
+            endpoint_url = s3_client.meta.endpoint_url
+            url_path = url.replace(endpoint_url, "").lstrip("/")
+            path = Path("/") / storage_code / url_path
```
sbidoul: I'm not entirely convinced the storage_code is that useful here, but to be seen when we integrate with the deployment pipeline.

lmignon: The mechanism is now generic. All the storages use the same rule into nginx. 
```
location ~ ^/fs_x_sendfile/(.*?)/(.*?)/(.*) {
        internal;
        proxy_pass http://myserver.com;
        set $url_protocol $1;
        set $url_host $2;
        set $url_path $3;
        set $url $url_protocol://$url_host/$url_path;
        proxy_pass $url$is_args$args;
        proxy_set_header Host $url_host;
        proxy_ssl_server_name on;
     
        # Optional: Forward original filename (if needed)
        proxy_set_header X-Original-URI $request_uri;
    }
```

fs_attachment_s3/models/fs_storage.py
```diff
@@ -0,0 +1,114 @@
...
+        fields.update(
+            {
+                "s3_uses_signed_url_for_x_accel_redirect": {},
+                "s3_signed_url_expiration": {},
+            }
+        )
+        return fields
+
+    @api.model
+    def _get_x_accel_redirect_path(self, attachment: IrAttachment):
```
sbidoul: Why not a method on ir.attachment?

fs_attachment_s3/readme/CONFIGURE.md
```diff
@@ -0,0 +1,57 @@
...
+involves setting up a location block in your web server configuration that
+matches the X-Accel-Redirect path and proxies the request to the S3 storage.
+
+For example, if you are using Nginx, you would add a location block like this:
+
+```nginx
+
+    location ~ ^/fs_x_accel_redirect/(.*?)/(.*?)/(.*) {
+        internal;
+        set $url_protocol $1;
```
sbidoul:
```suggestion
        set $url_scheme $1;
```

fs_attachment_s3/readme/CONFIGURE.md
```diff
@@ -0,0 +1,57 @@
...
+        set $url_host $2;
+        set $url_path $3;
+        set $url $url_protocol://$url_host/$url_path;
```
sbidoul:
```suggestion
        set $url $url_scheme://$url_host/$url_path;
```

fs_attachment_s3/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+    "name": "Fs Attachment S3",
+    "summary": """Store attachments into S3 complient filesystem""",
+    "version": "16.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "ACSONE SA/NV,Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/storage",
+    "depends": ["fs_attachment"],
+    "external_dependencies": {
+        "python": [
+            "s3fs",
```
sbidoul:
```suggestion
            "fsspec[s3]",
```

Would this work?

fs_attachment_s3/models/fs_storage.py
```diff
@@ -0,0 +1,114 @@
...
+        redirect_path = super()._get_x_accel_redirect_path(attachment)
+        fs = self.get_fs_by_code(attachment.fs_storage_code)
+        root_fs = self._get_root_filesystem(fs)
+        if hasattr(root_fs, "s3"):
+            redirect_path = self._get_s3_x_accel_redirect_path(attachment)
+        return redirect_path
```
sbidoul:
```suggestion
        fs = self.get_fs_by_code(attachment.fs_storage_code)
        root_fs = self._get_root_filesystem(fs)
        if hasattr(root_fs, "s3"):
            return self._get_s3_x_accel_redirect_path(attachment)
        return super()._get_x_accel_redirect_path(attachment)
```

fs_attachment_s3/models/fs_storage.py
lmignon: > Why not a method on ir.attachment?

I'm wondering the same thing.... so this method and the new one from the last commit must be moved on ir.attachment

lmignon:
@sbidoul Points that still need improvement:
* [x]  The code still lacks consistency. Odoo uses the term x-sendfile. In the code, I have a mix of method names. Sometimes you find x_accel_redirect, sometimes x_sendfile. I should only keep x_sendfile.
* [x]  Move the implementation of methods  ` _get_x_accel_redirect_path` and `_use_x_sendfile` from fs.storage to ir.attachment (remove this unnecessary delegation of the implemenatiion from ir.attachment to fs.storage).
* [ ] ~~Nice to have: to take advantage of the nginx cache, we should generate signed URLs for s3 with a long lifetime. We should also implement our own URL cache to return the same signed URL for the same object as long as it is not invalidated...~~

fs_attachment_s3/README.rst
```diff
@@ -0,0 +1,174 @@
...
+
+- ``{scheme}``: The URL scheme (http or https).
+- ``{host}``: The host of the S3 storage.
+- ``{path with query if any}``: The path to the file in the S3 storage,
+  including any query parameters. (Query parameters are set when the
+  ``s3_uses_signed_url_for_x_accel_redirect`` option is enabled.)
+
+In order to serve files using X-Accel-Redirect, you must ensure that
+your web server is configured to handle these headers correctly. This
+typically involves setting up a location block in your web server
```
sbidoul, line 90:
```suggestion
typically involves setting up an `internal <https://nginx.org/en/docs/http/ngx_http_core_module.html#internal>`_ location block in your web server
```

lmignon:
@sbidoul We need to adapt our deployements due to the last changes

sbidoul (approved):

lmignon:
Works in production for the last 2 weeks....

fs_attachment_s3/__manifest__.py
lmignon: AFAIK it should not work due to the way Odoo validates the declared python's external dependencies

PR #500, opened 2025-08-05
[FIX] storage_backend_sftp: Set paramiko version < 4.0.0 to ensure compatibility

baguenth:
On Aug4 a new version of Paramiko was released, which is not compatible with storage_backend_sftp.

https://pypi.org/project/paramiko/#history

Issue: 

ImportError: cannot import name 'DSSKey' from 'paramiko' 

Pinning the version to <4.0.0 should fix the issue for now, until someone took a closer look at the compatibly issue.

### DIFF
--- a/storage_backend_sftp/components/sftp_adapter.py
+++ b/storage_backend_sftp/components/sftp_adapter.py
@@ -32,12 +32,18 @@ def sftp_mkdirs(client, path, mode=511):
 
 
 def load_ssh_key(ssh_key_buffer):
-    for pkey_class in (
+    # Build list of supported key classes.
+    # Conditionally including DSSKey for backward compatibility with older
+    # versions of paramiko
+    pkey_classes = [
         paramiko.RSAKey,
-        paramiko.DSSKey,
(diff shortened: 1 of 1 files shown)


baguenth:
Hey @lmignon, I guess this is urgent ;) 

Potentially also other versions of this module are affected.

lmignon:
@baguenth Sorry but I no more use the storage_* addons... IMO it's always better to modify the code to make it compatible with the latest version of an external dependency if possible.  The removal of the support of DSSKey seems to be legit... https://github.com/paramiko/paramiko/issues/973

baguenth:
Hey @lmignon,

thank you for your feedback. I have updated the pull request to be more future proof.

Since you are not using the module anymore, can you may tell me who is the official maintainer?

baguenth:
Hey @simahawk, @etobella 

seems like @lmignon doesn't act as maintainer for this repo anymore. Can you may take a look? :)

lmignon (approved):

etobella (approved):

PR #514, opened 2025-09-17
[18.0][IMP] fs_storage: replace {db_name} in directory_path, fs_attachment and fs_attachment_s3: adapt to change

vincent-hatakeyama:
In multi database mode, there was no way to avoid filename collisions when a configuration file is used to define the fs_storage.storage to store attachments.

Port of #379

### DIFF
--- a/fs_attachment/models/fs_storage.py
+++ b/fs_attachment/models/fs_storage.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 
 from odoo import _, api, fields, models, tools
 from odoo.exceptions import ValidationError
+from odoo.tools import SQL
 from odoo.tools.safe_eval import const_eval
 
 from .ir_attachment import IrAttachment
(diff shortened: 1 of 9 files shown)


lmignon:
@vincent-hatakeyama Is the fs_attachment addon or others addons impacted by this change?

vincent-hatakeyama:
Yes, fs_attachment is impacted and changed by this PR.

Modules using directory_path (on fs.storage) should use `get_directory_path()` instead.

lmignon:
> directory_path

so the new addon `fs_attachment_s3` is also impacted

vincent-hatakeyama:
I’ve just changed fs_attachment_s3 too.

lmignon (approved):

lmignon:
ping @AnizR

AnizR (approved):
Code LGTM :+1: but, I didn't test it

PR #517, opened 2025-09-21
[18.0][IMP] fs_folder: A few UX improvements

jguenat:
Hello,

I watched the great talk about this new set of modules during OCA days.

I tried it with WebDAV (Nextcloud) and found some things that could be improved in my opinion.

- Add two buttons to unlink the folder and delete the folder
- Better manage an exception happening when fsspec does not find the folder on the remote.

Great set of modules I find it so much more comfortable to manage documents outside of Odoo with a dedicated software.

### DIFF
--- a/fs_folder/controllers/fs_folder_controller.py
+++ b/fs_folder/controllers/fs_folder_controller.py
@@ -140,3 +140,28 @@ class FsFolderController(http.Controller):
         request.env["fs.folder.field.web.api"].initialize_field_value(
             res_id, res_model, field_name
         )
+
+    @http.route(
+        "/fs_folder/unlink_folder/<string:res_model>/<int:res_id>/<string:field_name>",
+        type="json",
+        auth="user",
+        methods=["POST"],
+    )
+    def unlink_folder(self, res_id, res_model, field_name):
+        request.env["fs.folder.field.web.api"].remove_field_value(
+            res_id, res_model, field_name
+        )
+
+    @http.route(
+        "/fs_folder/delete_folder/<string:res_model>/<int:res_id>/<string:field_name>",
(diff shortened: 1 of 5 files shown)


lmignon (approved):
Thank you for this proposal @jguenat 
Adding the option to delete the directory created on the external file system could indeed be useful. I have often wondered whether or not to implement such a feature, given the consequences it may have. But in the end, there is no more risk than allowing the deletion of a subdirectory. The unlink feature is also useful in cases where a directory has been deleted on the external file system but is still referenced in Odoo. This allows the reference to be cleaned up without having to udate the database manually.

lmignon:
a little print-screen with the new buttons...
image

jguenat:
> Thank you for this proposal @jguenat Adding the option to delete the directory created on the external file system could indeed be useful. I have often wondered whether or not to implement such a feature, given the consequences it may have. But in the end, there is no more risk than allowing the deletion of a subdirectory. The unlink feature is also useful in cases where a directory has been deleted on the external file system but is still referenced in Odoo. This allows the reference to be cleaned up without having to udate the database manually.

That was also my reasoning.
The user can already delete all the folder files and sub-directories one by one. Also it's very easy to create a new root folder I think it's reasonable to allow the user to delete it (with a warning popup).
For the unlink yes we do not control what happens on the external file system so the case where the reference in Odoo is not valid might happen relatively often.

Thanks for the screenshot !

lmignon:
ping @pierre-halleux any comment?

pierre-halleux (approved):

pierre-halleux:
This is a really good idea. The unlink in particular will be very useful.

PR #520, opened 2025-09-28
[FIX] fs_attachment: use correct method call with context

kobros-tech:
get_default_storage_code_for_attachments method needs to be called with context otherwise it will return the default odoo code which is "file".

### DIFF
--- a/fs_attachment/models/ir_attachment.py
+++ b/fs_attachment/models/ir_attachment.py
@@ -254,7 +254,14 @@ class IrAttachment(models.Model):
     @api.model
     def _storage(self):
         # We check if a filesystem storage is configured for attachments
-        storage = self.env["fs.storage"].get_default_storage_code_for_attachments()
+        storage = (
+            self.env["fs.storage"]
+            .with_context(
+                attachment_res_model=self.res_model,
+                attachment_res_field=self.res_field,
+            )
+            .get_default_storage_code_for_attachments()
+        )
         if not storage:
             # If not, we use the default storage configured into odoo
             storage = super()._storage()


kobros-tech:
@lmarion-source 

This issue is critical and there is other one behind it, for current issue we suffer from saving product images inside odoo filestore instead of s3 one unless we set code="file"

There is an issue behind that which is that if the product image is saved correctly in s3 using code="s3" we can not see the image in the backend webclient for odoo although we can access the image url for that product publically.

kobros-tech:
https://github.com/user-attachments/assets/13da735a-eeb5-4107-9cfe-d9361b6c8d83

kobros-tech:
image

fs_attachment/models/ir_attachment.py
```diff
@@ -254,7 +254,14 @@ def _get_datas_related_values(self, data, mimetype):
     @api.model
     def _storage(self):
         # We check if a filesystem storage is configured for attachments
-        storage = self.env["fs.storage"].get_default_storage_code_for_attachments()
+        storage = (
+            self.env["fs.storage"]
+            .with_context(
+                attachment_res_model=self.res_model,
+                attachment_res_field=self.res_field,
```
lmignon, line 261: This method is an `api.model` method. Therefore, res_model and res_field are always False when called from the model. It's difficult to know what problem you're encountering. On my end, it's important to note that we use external file systems (mainly s3) to store our attachments by default without any issues with the code as it stands (which doesn't mean that a regression couldn't have been introduced).

lmignon (changes requested):

fs_attachment/models/ir_attachment.py
kobros-tech: the issue is that the context keys are falsy if you call the method without passing them.

I use s3 too and the only code that allows the attachment to be saved in external fs is file meaning the attachment will be saved inside odoo fs not in the external fs.

I add the two necessary contexts as calling this method without passing them has no real effect.

kobros-tech: they are not falsy as for product images, the passed in model is fs.product.image and the fields are passed in too.

If you suspect my issue I can show the log and passed in context values before and after my fix.

before my fix no attachment is saved in s3 fs but they are saved unfortunately in odoofs, after my fix attachments are saved in s3

lmignon: I didn't say you are wrong. I don't understand why in may case with the same addons installed, everything is working as expected. Moreover the method you modify is designed to be called on the model not on a recordset... 
As you can see, your change also introduce regressions into the tests.

kobros-tech: let me continue debugging I may find a result pleasing everybody.

lmignon: Maybe but take into consideration that the flow with fs.product.multi.image is also tested by unittest. So before trying to modify the code, try to reproduce the bug with a unittest.

lmignon: in your configuration you muse also add fs.image as saved into your s3 backend. If the image is linked to the fs.product.image, it's an fs.image record.

lmignon:
@kobros-tech I tested on runboat and everything is working as expected.

I configured a specific backend for fs.product.image and fs.image with the code fsimg.

image

I added an image on the product and linked an other one so I created a fs.product.image record and a fs.image record.

A created a server action searching for  fs.product.image and fs.image records and raising a UserError with the fs_filename of the attachment linked to both models.

image

The registered the ation on the attachment records and launched it..

The result show me that the files are stored into the expected storage.

image

kobros-tech:
thanks, it means I should follow the same steps.

let me do it in a few, but there must me some descriptive docs to guide users.

I may use this PR for that purpose if no coding is required.

lmignon:
Everyone is welcome to help improve the documentation. I have already put a lot of effort into it and into implementing all these modules....
This PR must be closed if there are no bugs.

kobros-tech:
After adding fs.image record I can see the correct code is returned by the method without using context, I am still not aware of everything here but I would prefer if using one model is enough without referring to liked source even by coding to simplify the implementation.

We are still facing another issue for displaying the images inside the webclient, do you know why I cannot see the image but I can access the image url?

image

image

image

lmignon:
> After adding fs.image record I can see the correct code is returned by the method without using context, I am still not aware of everything here but I would prefer if using one model is enough without referring to liked source even by coding to simplify the implementation.
> 
> We are still facing another issue for displaying the images inside the webclient, do you know why I cannot see the image but I can access the image url?
>  ....

I am sincerely sorry, but I am unable to provide full support for modules outside of my client base. Simply being notified of a problem reported as critical, for example, took me approximately half an hour of testing to conclude that there was no problem.

PR #522, opened 2025-10-02
[16.0][FIX] fs_storage: Resolve rooted_dir sub path is inside path virtually

### LINKED ISSUE #513: [16.0] fs_storage: issues with the rooted_dir_file_system
We are experiencing this intermittent issue where any uploads or reads from a configured `fs.storage` with `s3fs` are raising `FileNotFoundError` anytime `os.getcwd()` is called.

## Module
fs_storage

## Describe the bug
Mostly, anytime you try to see an image, upload an attachment, etc. if Odoo has to "contact" the external S3 bucket, it will raise `FileNotFoundError` from `os.getcwd()`.

## To Reproduce
**Affected versions**:
16.0
Possible anything else with the `rooted_dir_file_system`

1. If you have the bug, just refresh the record. For example in a `project.task` I sent some images to my customer, if I refresh the task, the images won't load or load sometimes and other times not.

**Expected behavior**
Issue not raising.

**Additional context**
```
  File "/opt/odoo/OCB/odoo/http.py", line 2234, in __call__
    response = request._serve_db()
  File "/opt/odoo/OCB/odoo/http.py", line 1822, in _serve_db
    return service_model.retrying(self._serve_ir_http, self.env)
  File "/opt/odoo/OCB/odoo/service/model.py", line 152, in retrying
    result = func()
  File "/opt/odoo/OCB/odoo/http.py", line 1850, in _serve_ir_http
    response = self.dispatcher.dispatch(rule.endpoint, args)
  File "/opt/odoo/OCB/odoo/http.py", line 1964, in dispatch
    return self.request.registry['ir.http']._dispatch(endpoint)
  File "/opt/odoo/OCB/addons/website/models/ir_http.py", line 237, in _dispatch
    response = super()._dispatch(endpoint)
  File "/opt/odoo/OCB/odoo/addons/base/models/ir_http.py", line 154, in _dispatch
    result = endpoint(**request.params)
  File "/opt/odoo/OCB/odoo/http.py", line 761, in route_wrapper
    result = endpoint(self, *args, **params_ok)
  File "/opt/odoo/OCB/addons/web/controllers/binary.py", line 171, in content_image
    return stream.get_response(**send_file_kwargs)
  File "/opt/odoo/kernet/oca/fs_attachment/fs_stream.py", line 72, in get_response
    res = _send_file(f, **send_file_kwargs)
  File "/opt/odoo/OCB/odoo/tools/_vendor/send_file.py", line 104, in send_file
    if isinstance(path_or_file, (os.PathLike, str)) or hasattr(
  File "/opt/odoo/kernet/oca/fs_attachment/models/ir_attachment.py", line 1142, in __getattr__
    self.__enter__()
  File "/opt/odoo/kernet/oca/fs_attachment/models/ir_attachment.py", line 989, in __enter__
    self._file_open()
  File "/opt/odoo/kernet/oca/fs_attachment/models/ir_attachment.py", line 1009, in _file_open
    fs, _storage, fname = self.attachment._get_fs_parts()
  File "/opt/odoo/kernet/oca/fs_attachment/models/ir_attachment.py", line 594, in _get_fs_parts
    return self._fs_parse_store_fname(self.store_fname)
  File "/opt/odoo/kernet/oca/fs_attachment/models/ir_attachment.py", line 527, in _fs_parse_store_fname
    fs = self._get_fs_storage_for_code(storage_code)
  File "/opt/odoo/kernet/oca/fs_attachment/models/ir_attachment.py", line 511, in _get_fs_storage_for_code
    fs = self.env["fs.storage"].get_fs_by_code(code)
  File "<decorator-gen-142>", line 2, in get_fs_by_code
  File "/opt/odoo/OCB/odoo/tools/cache.py", line 90, in lookup
    value = d[key] = self.method(*args, **kwargs)
  File "/opt/odoo/kernet/oca/fs_storage/models/fs_storage.py", line 227, in get_fs_by_code
    fs = fs_storage.fs
  File "/opt/odoo/kernet/oca/fs_storage/models/fs_storage.py", line 323, in fs
    raise e
  File "/opt/odoo/kernet/oca/fs_storage/models/fs_storage.py", line 319, in fs
    self._check_connection(self.__fs, self.check_connection_method)
  File "/opt/odoo/kernet/oca/fs_storage/models/fs_storage.py", line 306, in _check_connection
    self._ls_check_connection(fs)
  File "/opt/odoo/kernet/oca/fs_storage/models/fs_storage.py", line 300, in _ls_check_connection
    fs.ls("", detail=False)
  File "/opt/odoo/odooenv/lib/python3.8/site-packages/fsspec/implementations/dirfs.py", line 254, in ls
    ret = self.fs.ls(self._join(path), detail=detail, **kwargs).copy()
  File "/opt/odoo/kernet/oca/fs_storage/rooted_dir_file_system.py", line 47, in _join
    path_posix = os.path.normpath(make_path_posix(path))
  File "/opt/odoo/odooenv/lib/python3.8/site-packages/fsspec/implementations/local.py", line 304, in make_path_posix
    return f"{os.getcwd()}/{path}"
FileNotFoundError: [Errno 2] No such file or directory
```

A successful read looks like this:
```
2025-09-16 09:30:15,024 660916 INFO odoo odoo.addons.fs_storage.rooted_dir_file_system: ********************************************************************************
2025-09-16 09:30:15,024 660916 INFO odoo odoo.addons.fs_storage.rooted_dir_file_system: rooted_dir path: <bucket-name>/<file-checksum>
2025-09-16 09:30:15,024 660916 INFO odoo odoo.addons.fs_storage.rooted_dir_file_system: rooted_dir self.path: <bucket-name>
2025-09-16 09:30:15,025 660916 INFO odoo odoo.addons.fs_storage.rooted_dir_file_system: rooted_dir os.getcwd(): /opt/odoo
2025-09-16 09:30:15,025 660916 INFO odoo odoo.addons.fs_storage.rooted_dir_file_system: ********************************************************************************
```

A failed read looks like this:
```
2025-09-16 09:30:14,956 660907 INFO odoo odoo.addons.fs_storage.rooted_dir_file_system: ********************************************************************************
2025-09-16 09:30:14,956 660907 INFO odoo odoo.addons.fs_storage.rooted_dir_file_system: rooted_dir path: <bucket-name>
2025-09-16 09:30:14,956 660907 INFO odoo odoo.addons.fs_storage.rooted_dir_file_system: rooted_dir self.path: <bucket-name>
2025-09-16 09:30:14,956 660907 ERROR odoo odoo.addons.fs_storage.rooted_dir_file_system: [Errno 2] No such file or directory
```

In our `/etc/systemd/system/odoo.service` we have specified the `WorkingDirectory`:
`WorkingDirectory=/opt/odoo`

Any clue or fixes for this?

hnavarro-kernet:
Before this fix, a FileNotFoundError would ocassionally arise when 'os.getcwd()' is called due to the rooted_dir _join method to check whether the sub path is part of the parent path.
This is probably caused by the number and/or combination of workers and threads and the transactions to the FS storage, having race issues.

After this commit, the sub path is checked if it is contained inside the parent path virtually via posixpath and pathlib modules avoiding 'os.getcwd()' calls and running into this possible race condition.

closes #513

### DIFF
--- a/fs_storage/rooted_dir_file_system.py
+++ b/fs_storage/rooted_dir_file_system.py
@@ -1,10 +1,10 @@
 # Copyright 2023 ACSONE SA/NV (https://www.acsone.eu).
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 
-import os
+import posixpath
+from pathlib import PurePosixPath
 
 from fsspec.implementations.dirfs import DirFileSystem
-from fsspec.implementations.local import make_path_posix
 from fsspec.registry import register_implementation
 
 
@@ -20,18 +20,22 @@ class RootedDirFileSystem(DirFileSystem):
     """
 
     def _join(self, path):
-        path = super()._join(path)
+        joined = super()._join(path)
         # Ensure that the path is a subpath of the root path by resolving
         # any relative paths.
-        # Since the path separator is not always the same on all systems,
-        # we need to normalize the path separator.
-        path_posix = os.path.normpath(make_path_posix(path))
-        root_posix = os.path.normpath(make_path_posix(self.path))
-        if not path_posix.startswith(root_posix):
+
+        root = PurePosixPath(self.path).as_posix()
+        rnorm = posixpath.normpath(root)
+
+        jnorm = posixpath.normpath(joined or ".")
+
+        if not (jnorm == rnorm or jnorm.startswith(rnorm + "/")):
             raise PermissionError(
-                "Path %s is not a subpath of the root path %s" % (path, self.path)
+                f"Path {path!r} resolves to {jnorm!r} which is outside "
+                f"the root path {rnorm!r}"
             )
-        return path
+
+        return joined
 
 
 register_implementation("rooted_dir", RootedDirFileSystem)


hnavarro-kernet:
- [x] Fix tests

hnavarro-kernet:
- Fixed an issue with the patch that created a duplicated candidate: `subdirectory/here/subdirectory/here`, although it would not report an error
- Changed the implementation which caused absolute path candidates to fail even if they were an expansion of the root path

Ready for review, tagging @ivs-cetmix as the requestor of the PR from the issue.

cyrilmanuel (approved):

ivs-cetmix:
Hi @hnavarro-kernet thank you for your contribution! Unfortunately I don't possess enough technical knowledge of this module to do a comprehensive review, however I would kindly ask @sbidoul or maybe even @pedrobaeza  (why not 😄 ) to assist here.

pedrobaeza:
Not using them.

sbidoul:
I'll leave that to @lmignon 

But to better understand, why do you mention `os.getcwd`?  I don't see that in the diff, so I don't understand the problem this PR is solving.

fs_storage/rooted_dir_file_system.py
```diff
@@ -20,18 +20,20 @@ class RootedDirFileSystem(DirFileSystem):
     """
 
     def _join(self, path):
-        path = super()._join(path)
-        # Ensure that the path is a subpath of the root path by resolving
-        # any relative paths.
```
lmignon, line 25: @hnavarro-kernet Thank you for the fix. Nevertheless can you keep the first 2 comment lines

lmignon (changes requested):

hnavarro-kernet:
> I'll leave that to @lmignon
> 
> But to better understand, why do you mention `os.getcwd`? I don't see that in the diff, so I don't understand the problem this PR is solving.

As seen on the traceback reported in https://github.com/OCA/storage/issues/513, the usage of make_path_posix in the original rooted_dir_file_system https://github.com/OCA/storage/blob/b50f46883368e64445dba9554d905edfacebec33/fs_storage/rooted_dir_file_system.py#L28-L29 is the root cause of the os.getcwd calls, here's the original implementation:
https://github.com/fsspec/filesystem_spec/blob/e12aa7571244f6695264c92c4867978fed5ad092/fsspec/implementations/local.py#L319

sbidoul:
Ah, I see it now, thanks!

lmignon (approved):

fs_storage/rooted_dir_file_system.py
```diff
@@ -20,18 +20,22 @@ class RootedDirFileSystem(DirFileSystem):
...
-        # Since the path separator is not always the same on all systems,
-        # we need to normalize the path separator.
-        path_posix = os.path.normpath(make_path_posix(path))
-        root_posix = os.path.normpath(make_path_posix(self.path))
-        if not path_posix.startswith(root_posix):
+
+        root = PurePosixPath(self.path).as_posix()
+        rnorm = posixpath.normpath(root)
+
+        jnorm = posixpath.normpath(joined or ".")
```
sbidoul, line 30: I'm not convinced by this. Is it guaranteed that `self.path` will parse correctly as a PurePosixPath? Is it guaranteed that `joined` will be handled correctly by `posixpath.normpath` (Looking at the [DirFileSystem implementation](https://filesystem-spec.readthedocs.io/en/latest/_modules/fsspec/implementations/dirfs.html), it could be a dict or a list, and there is no guarantee the separator will be /).

sbidoul (changes requested):
I'm not sure... since this is security sensitive I'm going to mark it as Request changes to be sure everything is crystal clear (I may totally be missing something, though).

hnavarro-kernet:
> I'm not sure... since this is security sensitive I'm going to mark it as Request changes to be sure everything is crystal clear (I may totally be missing something, though).

Yeah, understandable.
I might have come too quickly into a fix that works for us, since the issue was growing rapidly becoming more and more reproducible and problematic.

I'll be happy to keep reviewing and raising concerns once I'm back from PTO next week.

fs_storage/rooted_dir_file_system.py
lmignon: indeed, ... this code lack of tests...
Considering the remarks mentioned here, the solution should take into account the separator to detect the kind of path (Posix or Windows)..... @sbidoul  I may be wrong, but regardless of the type of the path argument, the result of the call to super will always be a string...  We could add a check before the added logic to make sure, but in my opinion, it's useless. 

``` python
def _join(self, path):
    joined = super()._join(path)
    if not isinstance(joined, str):
        return joined
    # Uses the rigth path abstraction according to the
    # path separator
    if self.sep == '/':
        PathClass = PurePosixPath
        normpath = posixpath.normpath
    elif self.sep == '\\':
        PathClass = PureWindowsPath
        normpath = ntpath.normpath
    else:
        raise ValueError(f"Unknown path separator: {self.sep!r}")

    root = PathClass(self.path)
    joined_path = PathClass(joined or '.')

    rnorm = normpath(str(root))
    jnorm = normpath(str(joined_path))

    if not (jnorm == rnorm or jnorm.startswith(rnorm + self.sep)):
        raise PermissionError(
            f"Path {path!r} resolves to {jnorm!r} which is outside "
            f"the root path {rnorm!r}"
        )

    return joined
```

sbidoul: Maybe. But since we see that normpath is involving `os.getcwd`, does it make sense at all to use it on paths that may have nothing to do with the local file system?

lmignon: `os.getcwd` was called by the `make_path_posix` from `fsspec.implementations.local` AFAIK, it's not the case with the methods from python core.

PR #534, opened 2025-11-15
[ADD] fs_attachment_s3_migration: enable S3 migration

Aldeigja:
Helper module that migrates existing attachments to S3.


fs_attachment_s3_migration/models/ir_attachment.py
```diff
@@ -0,0 +1,252 @@
...
+        """Build domain for attachments to migrate, excluding force-DB files.
+
+        Respects force_db_for_default_attachment_rules to keep assets and
+        small images in database for performance.
+        """
+        base = [
+            ("store_fname", "not like", f"{storage_code}://%"),
+            "|",
+            ("res_field", "=", False),
+            ("res_field", "!=", False),
```
ivs-cetmix: Why do we need such domain leaf? Cannot it be simply omitted?

fs_attachment_s3_migration/models/ir_attachment.py
```diff
@@ -0,0 +1,252 @@
...
+
+        Prefer a donor already migrated to the target storage, then a donor
+        stored in DB. A donor is any attachment with the same checksum whose
+        ``datas`` can be read non-empty at this moment.
+        """
+        checksum = attachment.checksum
+        if not checksum:
+            return None
+
+        rf_domain = ["|", ("res_field", "=", False), ("res_field", "!=", False)]
```
ivs-cetmix: Same question about domain here. If this is needed on purpose, then there should be an explicit comment explaining the reason behind it.

fs_attachment_s3_migration/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,7 @@
+This module lets users move existing `ir.attachment` files from the
+standard filestore or database into an Amazon S3-backed `fs.storage`, using a
+ wizard directly on the storage form.
+
```
ivs-cetmix, line 4: Migrations are run in background batches, skipping attachments that are already stored in S3 or must remain in PostgreSQL.
This allows to run the process repeatedly avoiding creating duplicates.

fs_attachment_s3_migration/readme/CONTRIBUTORS.md
```diff
@@ -0,0 +1,4 @@
+* Cetmix (cetmix.com)
+* Ivan Sokolov
+* George Smirnov
```
ivs-cetmix: Cetmix (cetmix.com)
- Ivan Sokolov
- George Smirnov

fs_attachment_s3_migration/data/queue_job_channel_data.xml
ivs-cetmix, line 1: This setting is neither documented nor explained, however it's a very important one.

fs_attachment_s3_migration/wizard/migration_wizard.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class S3MigrationWizard(models.TransientModel):
+    _name = "s3.migration.wizard"
+    _description = "Migrate existing attachments to S3"
+
+    storage_id = fields.Many2one("fs.storage", required=True)
+    storage_code = fields.Char(required=True)
+    batch_size = fields.Integer(default=500)
+    channel = fields.Char(default="root.s3_migration")
```
ivs-cetmix: I think this is not a good idea to hardcorde it this way, especially takin into account the fact that you are using a data file to add it.
You should assign the default values using a function that populates the `env.ref("fs_attachment_s3_migration.queue_channel_s3_migration")` value.

fs_attachment_s3_migration/wizard/migration_wizard.py
```diff
@@ -0,0 +1,53 @@
...
+    max_batches = fields.Integer(
+        string="Max Batches (per click)",
+        help="Limit number of batches to enqueue now. Leave 0 for unlimited.",
+        default=0,
+    )
+
+    @api.onchange("storage_id")
```
ivs-cetmix: I would recommend to avoid  using `@onchange` and use compared stored fields with `readonly=False` instead.
This will simplify the usage as part of automation including external calls.

fs_attachment_s3_migration/models/ir_attachment.py
```diff
@@ -0,0 +1,252 @@
...
+                    "Migration batch progress: %d/%d (%.1f%%) - storage: %s",
+                    processed,
+                    total,
+                    (processed / total) * 100,
+                    storage_code,
+                )
+
+        _logger.info(
+            "Completed batch migration: %d/%d attachments to storage %s (%d skipped). "
+            "Old files will be cleaned by garbage collection.",
```
ivs-cetmix: by the garbage collector

fs_attachment_s3_migration/models/ir_attachment.py
```diff
@@ -0,0 +1,252 @@
...
+            )
+
+        _logger.info(
+            "Starting migration enqueue for storage %s (batch_size=%d, max_batches=%s)",
+            storage_code,
+            batch_size,
+            max_batches or "unlimited",
+        )
+
+        while True:
```
ivs-cetmix, line 182: I still have some concerns regarding this approach.

Pros
----

 Can run multiple jobs in parallel -> faster

Cons
----- 

Jobs are created instantly using the current DB snapshot.  So as soon as the migration is started no further changes are taken into account.
Eg if some attachments were removed they will be processed anyway even if this is not needed any longer.
When run in parallel several batches can access the same physical files (same checksum). If we modify the data in one of those batches the modification will not be saved in the database until the transaction is committed. So we either need to commit these changes explicitly (questionable) or prepare the batches checksum-grouped so attachments with the same checksum are always in the same batch.

Conclusion
-----------

Personally I would prefer to have those batches enqueued one after one to avoid potential issues and also always have an up-to date database snapshot being used. However I may be missing something here, so let's research this part better.

fs_attachment_s3_migration/models/ir_attachment.py
```diff
@@ -0,0 +1,252 @@
...
+                    attachment.id,
+                    str(e),
+                )
+                skipped += 1
+                processed += 1
+                continue
+
+            # Avoid empty writes if source is temporarily unreadable
+            if attachment.file_size and not file_data:
+                resolved = self._s3_resolve_migration_bytes(attachment, storage_code)
```
ivs-cetmix: This leads to a query being run N times which significantly degrades the performance especially on databases with large number of attachments.
I would suggest to fetch the data for the entire batch (or even the entire db 🤔 ) using `read_group` and then iterate it.

ivs-cetmix (changes requested):

fs_attachment_s3_migration/models/ir_attachment.py
Aldeigja: Good catch! This is a required workaround for Odoo's `ir.attachment._search` method 
which automatically adds `('res_field', '=', False)` to any domain that doesn't 
already contain 'res_field'. Without this tautology, only attachments where 
`res_field=False` would be found, missing all field-linked attachments.

I'll add an explicit comment explaining this.

fs_attachment_s3_migration/models/ir_attachment.py
Aldeigja: Same reason as above.

fs_attachment_s3_migration/data/queue_job_channel_data.xml
Aldeigja: Good point. I'll add comment a comment in the XML explaining the channel purpose.

fs_attachment_s3_migration/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+    "author": "Odoo Community Association (OCA), Cetmix",
+    "website": "https://github.com/OCA/storage",
+    "depends": [
+        "queue_job",
+        "fs_attachment_s3",
+    ],
+    "data": [
+        "security/ir.model.access.csv",
+        "data/queue_job_channel_data.xml",
+        "views/fs_storage_view.xml",
```
tendil: change and rename the file to `fs_storage_views.xml`

fs_attachment_s3_migration/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+        "views/migration_wizard_views.xml",
+    ],
+    "installable": True,
```
tendil: default value; can be deleted

fs_attachment_s3_migration/__manifest__.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2025 Cetmix OU
```
tendil: It must be **2026**, since the module was not merged in 2025
(change this in all other files as well)

fs_attachment_s3_migration/readme/CONTRIBUTORS.md
```diff
@@ -0,0 +1,5 @@
+Cetmix (cetmix.com)
```
tendil: replace the file at:

* [Cetmix](https://cetmix.com/):
  * Ivan Sokolov
  * George Smirnov
  
Example - https://github.com/cetmix/oca-sale-workflow/blob/e07ce66d12f7916fa50b8351ba3f3bc07a176d90/sale_order_team_from_product/readme/CONTRIBUTORS.md?plain=1

fs_attachment_s3_migration/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,9 @@
+This module lets users move existing `ir.attachment` files from the
+standard filestore or database into an Amazon S3-backed `fs.storage`, using a
```
tendil: there is a mismatch between the documented workflow and the implementation

the DESCRIPTION states that attachments are migrated from the filestore or database, while `_s3_migration_domain()` explicitly excludes db_datas
please either support DB-stored attachments or adjust the documentation and tests so the module contract is unambiguous

fs_attachment_s3_migration/models/ir_attachment.py
```diff
@@ -0,0 +1,333 @@
...
+        for candidate in candidates:
+            try:
+                data = candidate.raw
+                if data:
+                    return data
+            except OSError:
+                continue
+        return None
+
+    def _upload_to_storage(self, fs, path, bin_data):
```
tendil: this bypasses the `fs_attachment` storage lifecycle
direct `fs.open()` skips `_storage_write_option()` and does not register the newly created file through `_fs_mark_for_gc()`

for S3 it also skips the ContentType option
please keep the checksum deduplication, but reuse the core storage helpers/hooks so transaction/GC and backend-specific behavior are preserved

fs_attachment already has a `_storage_file_write()` function that centralizes:
- _get_fs_path()
- _storage_write_option(fs)
- actual fs.open()
- _fs_mark_for_gc()

https://github.com/OCA/storage/blob/1c43e2b2509ef5f6dbc32f434c8d2b8e37f9f5d1/fs_attachment/models/ir_attachment.py#L398

fs_attachment_s3_migration/models/ir_attachment.py
```diff
@@ -0,0 +1,333 @@
...
+
+        _logger.info(
+            "Completed migration enqueue: %d attachments in %d batches for storage %s",
+            total_enqueued,
+            batches,
+            storage_code,
+        )
+        return total_enqueued
+
+    def s3_migrate_batch(self, storage_code):
```
tendil, line 325: we need protection against concurrent updates here

core `_force_storage_to_object_storage()` locks the `ir_attachment` row with `FOR UPDATE NOWAIT` before reading and migrating it
currently an attachment can change between reading its checksum/data and `_force_write_store_fname()`, causing an updated record to be repointed to an old blob

fs_attachment_s3_migration/models/ir_attachment.py
```diff
@@ -0,0 +1,333 @@
...
+
+        try:
+            with fs.open(path, "wb") as f:
+                f.write(bin_data)
+        except OSError as e:
+            _logger.error("Failed to write file %s: %s", path, e)
+            raise
+
+    @api.model
+    def s3_enqueue_migration(
```
tendil: this is a public method and can be called through RPC with arbitrary parameters
the wizard ACL does not protect it
since this operation changes ir.attachment storage, it needs a server-side authorization check similar to `fs_attachment.force_storage()`

https://github.com/OCA/storage/blob/1c43e2b2509ef5f6dbc32f434c8d2b8e37f9f5d1/fs_attachment/models/ir_attachment.py#L748

https://www.odoo.com/documentation/16.0/developer/reference/backend/security.html

fs_attachment_s3_migration/models/ir_attachment.py
```diff
@@ -0,0 +1,333 @@
...
+                self._upload_to_storage(fs, path, bin_data)
+            except Exception as e:
+                _logger.error(
+                    "Upload failed for %s: %s, skipping %d attachments",
+                    checksum,
+                    e,
+                    len(attachments),
+                )
+                skipped += len(attachments)
+                continue
```
tendil: this exception should not be swallowed, otherwise `queue_job` considers the batch successful and `max_retries` never takes effect
a transient S3 failure should fail/retry the job (for example through the appropriate retryable exception), rather than permanently skipping these attachments

this also applies to the broad exception surrounding `_get_binary_data_for_checksum()`

fs_attachment_s3_migration/wizard/migration_wizard.py
```diff
@@ -0,0 +1,93 @@
...
+class S3MigrationWizard(models.TransientModel):
+    _name = "s3.migration.wizard"
+    _description = "Migrate existing attachments to S3"
+
+    storage_id = fields.Many2one(
+        "fs.storage",
+        string="Target Storage",
+        required=True,
+    )
+    storage_code = fields.Char(
```
tendil, line 19: storage_code should not be an independently editable value from storage_id

currently a user can select storage `A`, manually enter the code of storage `B`, and run the migration to `B` using configuration derived from `A`

please derive the code from `storage_id` and validate the S3 protocol server-side

fs_attachment_s3_migration/models/ir_attachment.py
```diff
@@ -0,0 +1,333 @@
...
+            storage_code,
+            batch_size,
+            max_batches or "unlimited",
+        )
+
+        while True:
+            cursor_domain = domain + [("id", ">", last_id)]
+            ids = (
+                self.with_context(prefetch_fields=False)
+                .search(cursor_domain, limit=batch_size, order="id ASC")
```
tendil: the previous concurrency issue for identical checksums is still not fully addressed: grouping happens only inside each batch, while batches are built with `order="id ASC"`

therefore the same checksum can still be split across parallel jobs. This also contradicts `test_enqueue_orders_by_checksum`

i would prefer either checksum-aware batching or chained batches/fresh pagination
queue_job 16.0 has `split(..., chain=True)` and job graphs

fs_attachment_s3_migration/models/ir_attachment.py
```diff
@@ -0,0 +1,333 @@
...
+                .ids
+            )
+            if not ids:
+                break
+
+            self.browse(ids).with_delay(
```
tendil: starting the migration again before the previous jobs finish will enqueue the same attachments again because their `store_fname` has not changed yet

please either use a deterministic `identity_key` or otherwise prevent overlapping migrations

fs_attachment_s3_migration/models/ir_attachment.py
```diff
@@ -0,0 +1,333 @@
...
+            new_store_fname = f"{storage_code}://{path}"
+            fs_filename = path.split("/")[-1]
+
+            for att in attachments:
+                old_fname = att.store_fname
+                if old_fname and not old_fname.startswith(f"{storage_code}://"):
+                    self._mark_old_store_fname_for_gc(old_fname)
+                att._force_write_store_fname(new_store_fname)
+
+            attachments.write({"fs_filename": fs_filename})
```
tendil: this bypasses the standard `fs_filename` semantics

`fs_attachment` uses `_enforce_meaningful_storage_filename()` and respects `use_filename_obfuscation`; here the checksum filename is always stored

please reuse the core filename handling

fs_attachment_s3_migration/models/ir_attachment.py
```diff
@@ -0,0 +1,333 @@
...
+                _logger.warning(
+                    "Cannot read checksum %s: %s, skipping %d attachments",
+                    checksum,
+                    e,
+                    len(attachments),
+                )
+                skipped += len(attachments)
+                continue
+
+            if not bin_data:
```
tendil: `b""` is valid content for a zero-byte attachment, but it is currently treated as missing data and skipped

please distinguish None from an empty bytes payload here

fs_attachment_s3_migration/models/ir_attachment.py
```diff
@@ -0,0 +1,333 @@
...
+                fs.makedirs(dirname, exist_ok=True)
+            except OSError as e:
+                _logger.debug("Directory %s may already exist: %s", dirname, e)
+
+        expected_size = len(bin_data)
+
+        if fs.exists(path):
+            try:
+                existing_size = fs.info(path).get("size", -1)
+                if existing_size == expected_size:
```
tendil, line 155: this is not content verification: equal size does not guarantee equal content

for a checksum-addressed object this early return can accept a corrupted S3 object as valid

please either verify the content/checksum or avoid describing this as content verification

fs_attachment_s3_migration/models/ir_attachment.py
```diff
@@ -0,0 +1,333 @@
...
+
+        fs_storage = self.env["fs.storage"].sudo().get_by_code(storage_code)
+        if fs_storage and fs_storage.migration_use_storage_force_db_rules:
+            force_db_domain = self._s3_get_force_db_domain(storage_code)
+            if force_db_domain:
+                return AND([base, ["!"] + force_db_domain])
+        return base
+
+    @api.model
+    def _s3_get_force_db_domain(self, storage_code):
```
tendil, line 73: this duplicates the force-DB logic already available in `fs_attachment` (`_store_in_db_instead_of_object_storage_domain()` and related helpers), which creates a semantic-drift risk

can we reuse/extend the core API instead of maintaining a second implementation of the same rules?

fs_attachment already has:
- `_store_in_db_instead_of_object_storage_domain()`
- `_store_in_db_instead_of_object_storage()`
- `_get_storage_force_db_config()`

https://github.com/OCA/storage/blob/1c43e2b2509ef5f6dbc32f434c8d2b8e37f9f5d1/fs_attachment/models/ir_attachment.py#L155

fs_attachment_s3_migration/models/ir_attachment.py
```diff
@@ -0,0 +1,333 @@
...
+        optimize_path = fs_storage.optimizes_directory_path
+
+        force_db_rules = {}
+        if fs_storage.migration_use_storage_force_db_rules:
+            force_db_rules = self.env[
+                "fs.storage"
+            ].get_force_db_for_default_attachment_rules(storage_code)
+
+        checksum_groups = {}
+        for att in self.with_context(prefetch_fields=False):
```
tendil: `prefetch_fields=False` here causes unnecessary per-record reads for checksum and other scalar fields

core disables prefetch during migration only when browsing records one-by-one to read datas, avoiding loading all blobs into memory
for metadata grouping, the normal batch prefetch should be preserved


in the parent migration, `prefetch_fields=False` is indeed used, but the reason there is different: records are browsed one at a time so that accessing the data doesn't force the system to read the content for hundreds of attachments at once and consume RAM
this is explicitly described in the core comment

fs_attachment_s3_migration/tests/test_attachment_migration.py
```diff
@@ -0,0 +1,340 @@
...
+            total = self.Attachment.s3_enqueue_migration(
+                "test_s3",
+                batch_size=10,
+                max_batches=2,
+            )
+
+        # Should stop at 20 (2 batches × 10)
+        self.assertEqual(total, 20)
+
+    def test_enqueue_orders_by_checksum(self):
```
tendil: this test does not verify its stated behavior: it inspects neither the ordering nor the queued batch contents, so it passes even with the current `order="id ASC"` 

`queue_job` recommends using `trap_jobs()` to assert the jobs and arguments that were actually enqueued

fs_attachment_s3_migration/tests/test_attachment_migration.py
```diff
@@ -0,0 +1,340 @@
...
+        cls.Attachment = cls.env["ir.attachment"]
+
+    def test_migrate_batch_handles_empty_recordset(self):
+        """Test that empty recordset doesn't crash."""
+        empty = self.Attachment.browse([])
+
+        # Should return True without error
+        result = empty.s3_migrate_batch("test_s3")
+        # Empty recordset may return False if storage not found, or True
+        self.assertIn(result, [True, False])
```
tendil: this assertion does not establish the method contract because it accepts both possible boolean results

please create a valid storage fixture and assert the concrete expected behavior for an empty batch

fs_attachment_s3_migration/tests/test_attachment_migration.py
```diff
@@ -0,0 +1,340 @@
...
+            mock_instance = MagicMock()
+            mock_delayable.return_value = mock_instance
+
+            total = self.Attachment.s3_enqueue_migration(
+                "test_s3",
+                batch_size=10,
+            )
+
+        # Should return count >= 5 (our attachments + any existing ones)
+        self.assertGreaterEqual(total, 5)
```
tendil: the test should not depend on unrelated attachments in the test database

this >= assertion can hide a bug where the records created by the test were not selected at all

please isolate the fixtures and assert the exact queued IDs/jobs

fs_attachment_s3_migration/wizard/migration_wizard.py
```diff
@@ -0,0 +1,93 @@
...
+
+    @api.depends("storage_id")
+    def _compute_channel(self):
+        for wizard in self:
+            if wizard.storage_id and wizard.storage_id.migration_channel:
+                wizard.channel = wizard.storage_id.migration_channel
+            elif not wizard.channel:
+                wizard.channel = wizard._get_default_channel()
+
+    def action_confirm(self):
```
tendil, line 88: migration parameters should be validated server-side before enqueueing: the target storage must exist and be S3, batch_size must be > 0, and max_batches >= 0. These values are currently editable and directly affect the workload

for the channel, I would either use `Many2one("queue.job.channel")` or, at the very least, validate the `complete_name`

tendil (commented):
Overall, it's a great idea, but to get the approve, you need to:

- fix the comments or explain why the change won't be accepted;
- cover the entire code with tests so it passes the CodeCov check;
- add complete docstrings for all methods that describe their parameters and return types.

fs_attachment_s3_migration/models/ir_attachment.py
```diff
@@ -0,0 +1,411 @@
...
+            "fs.storage"
+        ].get_force_db_for_default_attachment_rules(storage_code)
+        if force_db_config:
+            _logger.info(
+                "Migration will exclude force-DB files: %s",
+                force_db_config,
+            )
+
+        rows = self.search_read(domain, ["checksum"], order="checksum, id")
+        batches = self._s3_pack_checksum_batches(rows, batch_size)
```
tendil, line 264: `batch_size` does not currently bound the memory footprint of the enqueue operation
`search_read()` first materializes all eligible attachments in rows, after which additional batches and delayables lists are built in memory

on a large database this can again result in excessive memory usage or a MemoryError. max_batches does not help either, since it is applied only after the complete `search_read()`

please keep the checksum-aware grouping, but process the records incrementally using cursor/keyset pagination instead of materializing the entire `ir.attachment` selection at once

fs_attachment_s3_migration/models/ir_attachment.py
```diff
@@ -0,0 +1,411 @@
...
+                    identity_key=self._s3_migration_identity_key(
+                        storage_code, batch_ids
+                    ),
+                )
+                .s3_migrate_batch(storage_code)
+            )
+            total_enqueued += len(batch_ids)
+
+        if delayables:
+            chain(*delayables).delay()
```
tendil, line 294: do we actually need a `chain()` here?

it makes all migration batches strictly sequential, so increasing the capacity of the root.s3_migration channel has no effect. This also contradicts `queue_job_channel_data.xml` (_This channel allows controlling concurrency of migration jobs._), which documents the channel as a way to control migration parallelism

more importantly, if one job in the chain fails, all dependent jobs remain in wait_dependencies. `_s3_has_pending_migration_jobs()` considers that state an active migration, so another migration run is blocked until someone manually resolves the failed graph

since checksum groups are no longer split across batches and attachment updates are additionally protected by row locks, I would enqueue the batches independently unless there is a specific functional requirement to serialize the entire migration

fs_attachment_s3_migration/models/ir_attachment.py
```diff
@@ -0,0 +1,411 @@
...
+        self.ensure_one()
+        try:
+            with self.env.cr.savepoint():
+                self.env.cr.execute(
+                    "SELECT id FROM ir_attachment WHERE id = %s FOR UPDATE NOWAIT",
+                    (self.id,),
+                    log_exceptions=False,
+                )
+            return True
+        except psycopg2.OperationalError:
```
tendil, line 319: we should not treat every `psycopg2.OperationalError` as a row-lock conflict here 

operationalError may be raised for other database failures as well, in which case the attachment would be silently skipped and the job could still complete successfully

for `FOR UPDATE NOWAIT`, please handle only `PostgreSQL 55P03 (lock_not_available)` and re-raise any other operational error. Odoo uses the same pattern for NOWAIT locks

tendil (commented):
Thanks for fixing the comments!
Everything looks great now, and the comments below will most likely be the last ones.

please also take a look at the comment about the **dockstring for all methods** - it hasn't been fixed yet

PR #542, opened 2025-12-03
[FIX] fs_attachment: allow conccurency and chunk the cleanup request …

abraham-mplus:
…to a batch


lmignon (commented):
Thank you for the proposal. Can you add a unitest to ensure the new code is tested. Thanks!

PR #567, opened 2026-03-17
[18.0][FIX] Fix tests due to change in Odoo upstream

lmignon:


fs_folder/tests/common.py
```diff
@@ -48,6 +48,11 @@ def cleanup_tempdir():
 
     def setUp(self):
         super().setUp()
```
CRogos, line 50: shouldn't the result be returned at the end like in storage_thumbnail?

CRogos (approved):
LGTM. The change themes to work, but I am not an expert in this area.

But shouldn't we make one commit for each module with a correct commit massage?
[FIX] <module>: ....

fs_folder/tests/common.py
lmignon: no result is expected when calling setUp.

lmignon:
> LGTM. The change themes to work, but I am not an expert in this area.
> 
> But shouldn't we make one commit for each module with a correct commit massage? [FIX] : ....

It's a transversal fix that will not be forward ported. It only changes tests.

fs_folder/tests/common.py
CRogos: I know, but you did it different in the last module.

fs_folder/tests/common.py
```diff
@@ -48,6 +48,11 @@ def cleanup_tempdir():
...
+        # Remove check_attrs cleanup if exists to avoid conflict with FakeModelLoader.
+        # since superClass uses it for its own puposes not relevant for our tests.
+        check_attrs = (getattr(self, "check_attrs", None), tuple(), {})
+        if check_attrs in self._cleanups:
```
divad1196, line 54: Defining `check_attrs` empty should do the same while being more explicit on the intent and more resilient to Odoo changes.

```python
class FsFolderTestCase(TransactionCase):
    # Remove tests that are incompatible with FakeModuleLoader
    def check_attrs(self):
        pass
```

Any reason to not do it this way ?

PR #575, opened 2026-03-23
[16.0][ADD] image_tag_environment: make server_environment an optional dependency

yankinmax:
PR is aimed to remove `server_environment` dependency from `image_tag` in the scope of this issue:


image_tag_environment/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+    "summary": "Server environment features for the Image Tag model",
+    "version": "16.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "ACSONE SA/NV,Akretion,Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/storage",
+    "depends": ["image_tag", "server_environment"],
+    "data": [
+        "views/image_tag.xml",
+    ],
+    "auto_install": True,
```
ivantodorovich, line 15: Is it enough to get it automatically installed on environments where `image_tag` was already instealled and it only gets an update?

ivantodorovich (approved):

image_tag_environment/__manifest__.py
yankinmax: I've added the migration script and it's safe enough. No issues in logs about fields removal or schema change.

ivantodorovich:
@etobella this is similar to the one you did for fs_storage https://github.com/OCA/storage/pull/627
WDYT?

image_tag/upgrades/16.0.1.0.2/post-update.py
```diff
@@ -0,0 +1,13 @@
...
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    module = env["ir.module.module"].search(
+        [("name", "=", "image_tag_environment"), ("state", "=", "uninstalled")]
+    )
+    if module:
+        module.button_install()
```
etobella, line 13: My experience also states that it is nice to get the data and add it to the right field. This happened to me and I would recommend you to do it this way 😉

etobella (approved):
For me OK, just review my comment if possible 😉

image_tag/upgrades/16.0.1.0.2/post-update.py
ivantodorovich: Which data? the one set in the environment?

The purpose of this script is to install the new split off module, so instances that were configured through the environment remain configurable through env -- by having the new `*_environment` module automatically installed

Users or sysadmins are expected to explicitly uninstall the `*_environment` module and re-configure their system.

To make that easier, we're working on this approach:
- https://github.com/OCA/server-env/pull/262
- https://github.com/OCA/server-env/pull/261

It's not done in this PR, though. At this stage we want at least to have the modules split.

etobella: yes, but it will give you problems when you want to remove the environment, also, it happened to me that I decided to remove the dependancy as it was not required, at that moment, everything failed as the uninstall hook did not pass.... I am just warning, it shouldn't be a big deal to handle in the migration too.

yankinmax: We don't need any uninstall hook here, because `image_tag_environment` doesn't actually override `_server_env_fields`

yankinmax:
@etobella for this split case we don't need any extra uninstall hooks or UserError exceptions. No field is namaged as server env field.

etobella:
then the user will loose the data as the field where it is stored will be deleted. that happened to me once in storage. not blocking for me as I am not using it anymore, but I was warning you

lmignon (approved):
IMO `image_tag_environment` will never be migrated and disapear... I'm wondering why i put a dependency on server_environment 🤔 🙄

PR #576, opened 2026-03-23
[16.0][ADD] storage_backend_environment: make server_environment an optional dependency

yankinmax:
PR is aimed to remove `server_environment` dependency from `storage_backend` in the scope of this issue:


storage_backend_environment/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+    "summary": "Use Server Environment feature to manage storage backends",
+    "version": "16.0.1.0.0",
+    "category": "Storage",
+    "website": "https://github.com/OCA/storage",
+    "author": " Akretion, Odoo Community Association (OCA)",
+    "license": "LGPL-3",
+    "development_status": "Production/Stable",
+    "installable": True,
+    "depends": ["storage_backend", "server_environment"],
+    "auto_install": True,
```
ivantodorovich, line 16: same questions than other PRs.

Is it enough to get it automatically installed on environments where `image_tag` was already instealled and it only gets an update? or do we need a migration script?

ivantodorovich (approved):

lmignon:
@yankinmax The issue is for V19. Here you are on a stable release. Do you plan to still use these storage addons in the future? You've new alternatives with the fs_* addons.

yankinmax:
> @yankinmax The issue is for V19. Here you are on a stable release. Do you plan to still use these storage addons in the future? You've new alternatives with the fs_* addons.

The idea was to adapt all the modules that have a direct `server_environment` dependency.
Ping @ivantodorovich @gurneyalex

PR #579, opened 2026-03-24
[19.0][ADD] fs_storage_environment: make server_environment an optional dependency

yankinmax:
Forward port of:
- [ ] https://github.com/OCA/storage/pull/578

PR is aimed to remove `server_environment` dependency from `fs_storage` in the scope of this issue:


yankinmax:
Hello @ivantodorovich 
I've checked this migration process on the Odoo v19 database having some server env variables defined.
In general the state before and after is preserved.
I attach the logs.
[server.log](https://github.com/user-attachments/files/26380207/server.log)

ivantodorovich:
@yankinmax I updated the script in the last commit.

Can you check on your end, to see if the tests you were executing still pass?
if you agree with this, I let you clean up the PR (squash commits, etc..) and port this to the other branches

yankinmax:
> @yankinmax I updated the script in the last commit.
> 
> Can you check on your end, to see if the tests you were executing still pass? if you agree with this, I let you clean up the PR (squash commits, etc..) and port this to the other branches

Hello @ivantodorovich 
I've checked your idea and it seems a nice tuning of migration process design we started to implememnt.
I've tested this approach on the real project data.
1. The fields are being renamed. No more removal. This is great.
2. WE don't care anymore about data loss, so my part of preserving protocol column is not needed anymore.
3. But, we still have newly created columns coming from the transient state when `server.env.mixin` is not yet applied. So, I've added a fixup commit to preserve `fs_storage` table columns state before and after migration. It has a safe exit if this is a brand new `fs_storage_environment` installation.


What I can't understand: why do we have such an error in the tests: 
```
ModuleNotFoundError: No module named 'openupgradelib'
```
I've never seen `openupgradelib` not available.

ivantodorovich:
AFAICS there was no data loss (2). What do you mean? The protocol column is configured through environment 

And we don't care about the created columns (3). That's normal. Same thing happens for example with ir_mail_server and mail_environment.

yankinmax:
> AFAICS there was no data loss (2). What do you mean? The protocol column is configured through environment
> 
> And we don't care about the created columns (3). That's normal. Same thing happens for example with ir_mail_server and mail_environment.

if we don't care about the columns being created in the `fs_storage` table other points are ok

ivantodorovich:
Yeap, IMO it's better without these additions. I'd like to KISS cause we'll need to use the same/similar script in all other  modules.

fs_storage/__manifest__.py
```diff
@@ -5,18 +5,18 @@
...
     "development_status": "Beta",
-    "depends": ["base", "base_sparse_field", "server_environment"],
+    "depends": ["base", "base_sparse_field"],
     "data": [
         "views/fs_storage_view.xml",
         "security/ir.model.access.csv",
         "wizards/fs_test_connection.xml",
     ],
-    "external_dependencies": {"python": ["fsspec>=2024.5.0"]},
+    "external_dependencies": {"python": ["fsspec>=2024.5.0", "openupgradelib>=3.6.0"]},
```
yankinmax, line 20: I had several times an issue with `openupgradelib` no available. I don't get why.

yankinmax:
I can't get the idea why tests fail.
Help is appreciated.
 ```
INFO odoo odoo.addons.fs_attachment.tests.test_fs_storage: Starting TestFsStorage.test_url_for_image_dir_optimized_and_not_obfuscated ... 
2026-04-07 10:05:02,626 338 INFO odoo odoo.addons.base.models.ir_attachment: filestore gc 2 checked, 2 removed 
2026-04-07 10:05:02,708 338 INFO odoo odoo.tests.suite: ====================================================================== 
2026-04-07 10:05:02,708 338 ERROR odoo odoo.tests.suite: ERROR: setUpClass (odoo.addons.fs_attachment.tests.test_stream.TestStream)
  File "/__w/storage/storage/fs_attachment/tests/test_stream.py", line 54, in setUpClass
    assert cls.attachment_binary.fs_filename
AssertionError
```

ivantodorovich:
Hmm, @yankinmax  debug locally the `_enforce_meaningful_storage_filename` and `_build_fs_filename` methods. It's possible that with the server_environment split, the underlying `fs.storage` records in `fs_attachment` tests have changed, and giving different results

PR #581, opened 2026-03-26
[18.0] storage_image_product: ease test setup for other modules

simahawk:
Having a mixin class allows to run the setup of the base data explicitly where needed especially when you have test classes with multiple inheritances.

After the Odoo tests apocalypse we cannot rely anymore on running Klass.setUpClass() w/o tons of attributes warning. Now you can inherit the mixin and do the setup properly.


simahawk:
Tested :heavy_check_mark:  Fast tracking :speedboat: 
/ocabot merge patch

PR #582, opened 2026-03-30
[18.0][ADD] image_tag_environment: make server_environment an optional dependency

yankinmax:
Forward-port of:
- [ ] https://github.com/OCA/storage/pull/575

PR is aimed to remove `server_environment` dependency from `image_tag` in the scope of this issue:


ivantodorovich (approved):

image_tag_environment/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+    "summary": "Server environment features for the Image Tag model",
+    "version": "18.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "ACSONE SA/NV,Akretion,Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/storage",
+    "depends": ["image_tag", "server_environment"],
+    "data": [
+        "views/image_tag.xml",
+    ],
+    "auto_install": True,
```
simahawk: as pointed out on #654 IMO these modules should not have auto_install enabled

simahawk (changes requested):

simahawk (approved):

PR #584, opened 2026-04-01
[18.0][IMP] fs_attachment: enhance stream download name handling based on mimetype

CRogos:
Fixes missing file extension when downloading with Microsoft Edge.

image

Alligns with: https://github.com/odoo/odoo/blob/18.0/odoo/addons/base/models/ir_binary.py#L148-L149


lmignon (approved):
Thank you for the fix @CRogos . I wonder why the extension isn't included in the name from the start 🤔

CRogos:
It also depends of the browser you've tested with. With Firefox the .pdf is added without this fix... Chrome I did not test.

CRogos:
@sbidoul could you do the second review?

leemannd (approved):

PR #585, opened 2026-04-01
[18.0][IMP] fs_attachment: simplify domain for attachment search by skipping res_field checks

CRogos:
@lmignon what do you think on this change? Shouldn't this bypass this check as intended?

https://github.com/odoo/odoo/blob/18.0/odoo/addons/base/models/ir_attachment.py#L531-L535

PS: I haven't tested it yet, therefor draft. Tests are green.

### DIFF
--- a/fs_attachment/models/fs_storage.py
+++ b/fs_attachment/models/fs_storage.py
@@ -310,19 +310,14 @@ class FsStorage(models.Model):
         in staging are done in a different directory and will  not impact the
         production.
         """
-        # The weird "res_field = False OR res_field != False" domain
-        # is required! It's because of an override of _search in ir.attachment
-        # which adds ('res_field', '=', False) when the domain does not
-        # contain 'res_field'.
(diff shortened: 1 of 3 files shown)


fs_attachment/models/fs_storage.py
```diff
@@ -310,19 +310,14 @@ def recompute_urls(self) -> None:
...
-            "|",
-            ("res_field", "=", False),
-            ("res_field", "!=", False),
         ]
-        attachments = self.env["ir.attachment"].search(domain)
+        attachments = (
+            self.env["ir.attachment"]
+            .with_context(skip_res_field_check=True)
+            .search(domain)
+        )
```
CRogos, line 320: Should be covered by this test: 

https://github.com/oca/storage/blob/18.0/fs_attachment/tests/test_fs_storage.py#L245-L279

lmignon:
LGTM bu should be tested

FrancoMaxime (commented):
LGTM: code review, i haven't a 18 version to functionally test the PR

PR #562, opened 2026-03-11
[19.0][MIG] fs_folder

sbejaoui:

PR #591, opened 2026-04-14
[19.0][MIG] fs_folder: Migration to 19.0

quoc-pn:


lmignon:
@quoc-pn Thank your for your proposal. Nevertheless an PR already exists for the migration. Can you take a look at https://github.com/OCA/storage/pull/562 to help on the migration process? Thank you in advance.

quoc-pn:
@lmignon Thank you for your review.
I reviewed the existing PR, and it doesn't meet the OCA migration standard. 
Since the author has not responded, I need to open another PR to resolve my current use case.

PR #593, opened 2026-04-14
FIX fs_storage: typo in options

bealdav:
In my experience it works like that for sftp or ftp

@hparfr @lmignon


lmignon (approved):
@bealdav I it works like that, it's ok for me... ;-)
It's inline with the documentation

hparfr (approved):

PR #605, opened 2026-05-19
[19.0][IMP] fs_folder: Increase preview iframe z-index

quoc-pn:
### Issue
Currently, the status bar in the form view has a `z-index` of `1020`, which is higher than the preview iframe’s `z-index` of `1000`.

As a result, the status bar is displayed above the preview iframe.

image

### Solution
Increase the preview iframe’s `z-index` to `9999` to ensure it is displayed correctly above the form view status bar.

### DIFF
--- a/fs_folder/static/src/preview_iframe/preview_iframe.scss
+++ b/fs_folder/static/src/preview_iframe/preview_iframe.scss
@@ -8,7 +8,7 @@
     display: flex;
     justify-content: center;
     align-items: center;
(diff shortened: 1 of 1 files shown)


lmignon (approved):
Thank you for the fix @quoc-pn

lmignon:
@quoc-pn Your PR targets 19.0 but your branch is not based on the migrated one https://github.com/OCA/storage/pull/591

PR #603, opened 2026-05-13
[FIX] test-requirements: pin max supported version for fsspec deps

simahawk:
Tries to fix #602 

There were 3 modules w/ 3 different versions of fsspec + no cap on test deps.

Collateral fix: remove obsolete `setup` folder.

PR #606, opened 2026-05-20
[18.0][IMP] fs_storage: add _move_file helper

ArnauCForgeFlow:
Add a generic `_move_file` helper on `fs.storage`: checks that the source file exists and schedules `fs.move` as a post-commit hook so the storage stays in sync with the DB transaction.

### DIFF
--- a/fs_storage/models/fs_storage.py
+++ b/fs_storage/models/fs_storage.py
@@ -10,6 +10,7 @@ import logging
 import os.path
 import re
 import warnings
+from pathlib import PurePath
 from typing import AnyStr
 
 import fsspec
@@ -819,3 +820,12 @@ class FSStorage(models.Model):
         while hasattr(fs, "fs"):
             fs = fs.fs
         return fs
+
+    def _move_file(self, from_dir_str, to_dir_str, filename):
+        self.ensure_one()
+        src = (PurePath(from_dir_str) / filename).as_posix()
+        if not self.fs.exists(src):
+            return False
+        dst = (PurePath(to_dir_str) / filename).as_posix()
+        self.env.cr.postcommit.add(functools.partial(self.fs.move, src, dst))
+        return True
--- a/fs_storage/tests/test_fs_storage.py
+++ b/fs_storage/tests/test_fs_storage.py
@@ -304,3 +304,22 @@ class TestFSStorage(TestFSStorageCase):
             safe_eval.safe_eval(
                 "env['fs.storage'].search([]).unlink()", {"env": self.env}
             )
+
+    def test_move_file(self):
+        src_dir = "src"
+        dst_dir = "dst"
+        self.backend.fs.makedirs(src_dir, exist_ok=True)
+        self.backend.fs.makedirs(dst_dir, exist_ok=True)
+        self._create_file(self.backend, f"{src_dir}/{self.filename}", self.filedata)
+
+        res = self.backend._move_file(src_dir, dst_dir, self.filename)
+        self.assertTrue(res)
+        # The actual move runs on post-commit, so until then the file
+        # must still live in the source directory.
+        self.assertTrue(self.backend.fs.exists(f"{src_dir}/{self.filename}"))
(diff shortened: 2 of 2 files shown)


simahawk:
Build fixed here https://github.com/OCA/storage/pull/603

ArnauCForgeFlow:
Rebased @simahawk

LoisRForgeFlow (approved):
LGTM

fs_storage/models/fs_storage.py
```diff
@@ -819,3 +820,19 @@ def _get_root_filesystem(self, fs=None):
...
         return fs
+
+    def _move_file(self, from_dir_str, to_dir_str, filename):
+        self.ensure_one()
+        from_dir = PurePath(from_dir_str)
+        to_dir = PurePath(to_dir_str)
+        if not self.fs.exists(from_dir.as_posix()):
+            return False
+        files = [
+            os.path.basename(f) for f in self.fs.ls(from_dir.as_posix(), detail=False)
```
simahawk: If the dir is big don't we risk that this `ls` takes a lot of time?
Wouldn't be better to check if `dir/filename` exists or just let if fail it doesn't? 🤔

simahawk (changes requested):
would be nice to have some test cov

ArnauCForgeFlow:
@simahawk added tests

simahawk (approved):

ArnauCForgeFlow:
can we merge? @simahawk

fs_storage/models/fs_storage.py
```diff
@@ -819,3 +820,12 @@ def _get_root_filesystem(self, fs=None):
...
             fs = fs.fs
         return fs
+
+    def _move_file(self, from_dir_str, to_dir_str, filename):
+        self.ensure_one()
+        src = (PurePath(from_dir_str) / filename).as_posix()
+        if not self.fs.exists(src):
+            return False
+        dst = (PurePath(to_dir_str) / filename).as_posix()
+        self.env.cr.postcommit.add(functools.partial(self.fs.move, src, dst))
```
lmignon, line 830: The intent of the fs_storage library is to provide access to an fsspec proxy client instance based on a configured storage backend. The original intention was not to build an additional API layer on top of the fsspec API itself.

To me, it feels strange that calling the `_move_file` method would not actually move the file by the end of the method call.

If the goal is to address a transactional consistency issue — ensuring that modifications made to the external filesystem remain consistent with the Odoo transaction — then the approach should probably be different, and significantly more complex. (we should rely on the transactional layer of the fsspec library requiring that the different client stubs supports properly the caching mechanism)

In this specific case, you could simply move the file directly within the method and register a rollback hook to move it back if the transaction fails.

So, to be honest, I’m a bit confused by this proposal.

lmignon (changes requested):

fs_storage/models/fs_storage.py
simahawk: Ciao @lmignon :)
The fs_storage module itself is already providing a wrapper for that lib.
In this case, we want to make sure that IF for whatever reason the transaction fails on odoo, the file remains where it is.
When you say " the approach should probably be different, and significantly more complex. " do you have anything in mind?
Anyway, we could remove the postcommit part and offer an hook to customize this behavior elsewhere.

lmignon: I've not investigated further in details but fsspecs comes with a concept of transaction https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.spec.AbstractFileSystem.transaction Maybe it doesn't apply to such a kind of operation and I've the feeling it introduce too much complexity.... 

Regarding the current proposal, I see 2 problems:
* The call to the method has no direct effect. If the caller expect that the file has been moved as a result to the call to the method, it's not the case. (This should be reflected somewhere into the doc, the name of the method or ...)
* Since the move is done after the transaction, if it fails, you are also in a incoherent state (yes we are not into a distributed transaction system....)

PR #608, opened 2026-05-20
[16.0][IMP] fs_file_X: fs_file based addons are no more Alpha

lmignon:
They work in production for more than 3 years now...


lmignon:
ping @sebastienbeau

PR #610, opened 2026-05-26
[16.0][FIX] fs_attachment: paginate autovacuum GC loop to bound worker memory

lmignon:
supersedes #597


rrebollo (commented):
Once merged, we need to forward-port this to higher versions.

lmignon:
this one run in production without issue for the last couple of weeks... merging....

PR #614, opened 2026-05-26
[16.0][IMP] fs_storage: replace {db_name} in directory_path, fs_attachment …

lmignon:
…and fs_attachment_s3: adapt to change

In multi database mode, there was no way to avoid filename collisions when a configuration file is used to define the fs_storage.storage to store attachments.

backport of #379

### DIFF
--- a/fs_attachment/models/fs_storage.py
+++ b/fs_attachment/models/fs_storage.py
@@ -3,9 +3,10 @@
 
 from __future__ import annotations
 
+from psycopg2.extensions import AsIs
+
 from odoo import _, api, fields, models, tools
 from odoo.exceptions import ValidationError
-from odoo.tools import SQL
 from odoo.tools.safe_eval import const_eval
 
 from .ir_attachment import IrAttachment
@@ -497,7 +498,7 @@ class FsStorage(models.Model):
         # This is needed when {db_name} is used in directory path.
         # Due to the use of server_environment, there is no directory_path column to
         # filter so recompute all.
-        self.env.cr.execute(SQL("SELECT id FROM %s", SQL.identifier(self._table)))
+        self.env.cr.execute("SELECT id FROM %s", (AsIs(self._table),))
         records = self.browse(row[0] for row in self.env.cr.fetchall())
         if records:
             self.env.add_to_compute(self._fields["base_url_for_files"], records)
--- a/fs_attachment/models/fs_storage.py
+++ b/fs_attachment/models/fs_storage.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 
 from odoo import _, api, fields, models, tools
 from odoo.exceptions import ValidationError
+from odoo.tools import SQL
 from odoo.tools.safe_eval import const_eval
 
 from .ir_attachment import IrAttachment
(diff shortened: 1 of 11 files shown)


etobella:
@lmignon This code is giving problems if you have added other fsspec options. This should be removed.

etobella:
After a careful review, the problem comes from an incompatibilities between:

1- Adding new fsspec types on a module
2- Usage of server-env (reviews the options, as it is not added yet, it fails)
3- This code

If we remove one fo this 3 conditions, it works, however, as 1 and 2 were here before, I think we need to fix this one. the error comes from the computing in _setup_complete, that is executed before all modules are loaded. So, if you added a fsspec definition on a module, everything starts to fail...

lmignon:
Thank you Enric for the report. As I hadn’t received any feedback regarding a potential issue and hadn’t encountered such a problem myself, I didn’t realise that this code could cause a problem. Unfortunately, I don’t really understand how to reproduce the problem described...

etobella:
Declare a new module that creates a new kind of fsspec

```python
from fsspec.registry import register_implementation
from fsspec.spec import AbstractBufferedFile, AbstractFileSystem


class DemoFileSystem(AbstractFileSystem):
    """An integrated system to send data."""

    protocol = "demo"


register_implementation("demo",DemoFileSystem)
```

And add this spec in the models


```python
from odoo import api, models

class FsStorage(models.Model):

    _inherit = "fs.storage"

    @api.model
    def _get_protocols(self) -> list[tuple[str, str]]:
        protocol = super()._get_protocols()
        protocol.append(("demo", "Demo"))
        return protocol

    @api.model
    def _get_options_protocol(self) -> list[tuple[str, str]]:
        protocol = super()._get_options_protocol()
        protocol.append(("demo", "Demo"))
        return protocol

```

Create a fs storage with this kind and you will no longer be able to upgrade.

THis is failing now and before it worked.

The problem comes from recomputing here some data on upgrade. The main issue is that the protocol is not defined yet on fs_attachment and that is giving an error.

lmignon:
> Declare a new module that creates a new kind of fsspec
> 
> ```python
> from fsspec.registry import register_implementation
> from fsspec.spec import AbstractBufferedFile, AbstractFileSystem
> 
> 
> class DemoFileSystem(AbstractFileSystem):
>     """An integrated system to send data."""
> 
>     protocol = "demo"
> 
> 
> register_implementation("demo",DemoFileSystem)
> ```
> 
> And add this spec in the models
> 
> ```python
> from odoo import api, models
> 
> class FsStorage(models.Model):
> 
>     _inherit = "fs.storage"
> 
>     @api.model
>     def _get_protocols(self) -> list[tuple[str, str]]:
>         protocol = super()._get_protocols()
>         protocol.append(("demo", "Demo"))
>         return protocol
> 
>     @api.model
>     def _get_options_protocol(self) -> list[tuple[str, str]]:
>         protocol = super()._get_options_protocol()
>         protocol.append(("demo", "Demo"))
>         return protocol
> ```
> 
> Create a fs storage with this kind and you will no longer be able to upgrade.
> 
> THis is failing now and before it worked.
> 
> The problem comes from recomputing here some data on upgrade. The main issue is that the protocol is not defined yet on fs_attachment and that is giving an error.

Ouchh a new fsspec filesystem adapter should be developed as a pure python lib as I did it for msgraphfs. I should do it for the RootedDirFileSystem and OdooFileSystem classes and remove them from the fs_folder addon. A temp solution could be, if you manage your addon path by hand, to put your filesystem into a dedicated addon without dependency and set this addon in first position of your addon path.  WDT?

etobella:
Easy to say now, but it was allowed and now it doesn't. Also, the problem comes from the fact of forcing server_environment when it should not be required...

etobella:
Also, there is nothing in the documentation that specifies that it should be done in a library...

etobella:
And just to give another reason for removing the problem here is that we should not compute the field like this, what happens if I made a modification on such a function? We shouldn't enforce it this way.

lmignon:
I’m open to any suggestions for bug fixes. Personally, I don’t use this feature. The initial pull request remained open for a very long time before being merged and later back/forward ported. Unfortunately I hadn’t anticipated this kind of issue.

lmignon:
@etobella I'll propose a Fix.

PR #599, opened 2026-04-28
[17.0][IMP] fs_attachment_s3: Optimize clean up storages using mass deletion

peluko00:
Optimize clean up storages using mass deletion with S3
cc https://github.com/APSL 38655
@mpascuall @BernatObrador @ppyczko @shirashi3771 @palomagrc93 @miquelalzanillas @LlucSG please review

PR #615, opened 2026-05-26
[16.0] [IMP] fs_attachment_s3: Optimize clean up storages using mass deletion

lmignon:
backport and improves #599

* [x] requires #610


lmignon:
this one run in production without issue... merging....

PR #626, opened 2026-05-29
[19.0][IMP] fs_folder: Disable FsFolderItem clicks while processing

quoc-pn:
## Summary
Fix race conditions in FsFolder by preventing overlapping FsFolderItem actions from rapid clicks. 

image

## What changed 
- Added a single in-flight lock in FsFolder state. 
- Routed item actions through the parent lock so only one action runs at a time. 
- Released the lock in all cases (success or failure). 

## Result 
- No duplicate item actions from double/rapid clicks. 
- View interaction is restored reliably after each action.


lmignon (commented):
Thank you @quoc-pn Fot the fix. I'm no a JS specialist.... so I'll take time to analyse your change and see if it's the right pattern (at least for my internal knowledge)

PR #628, opened 2026-06-03
[16.0][FIX] fs_attachment: Delay base_url_for_files recompute at end of reg…

lmignon:
An attempt to fix problem reported in #614

### DIFF
--- a/fs_attachment/models/fs_storage.py
+++ b/fs_attachment/models/fs_storage.py
@@ -3,8 +3,6 @@
 
 from __future__ import annotations
 
-from psycopg2.extensions import AsIs
-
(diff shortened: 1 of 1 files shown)


lmignon:
@etobella Do you think this could fix your problem or do we have to completely revert changes from #614 (but they are also present into 17, 18 and 19) 
From 20 we will remove the dependency on server_environement.

pedrobaeza:
Sorry, but I think the dependency should be removed ASAP, as it's not adding value, but conditioning things and provoking incompatibilities in systems not using them, and you can see that the migration script will handle the case where you need it.

PR #632, opened 2026-06-09
[19.0][FIX] fs_attachment: error on http stream

### LINKED ISSUE #631: [19.0][fs_storage] Error 500 on direct attachment streaming
Method _to_http_stream not overridden for attachments, it causes causes FileNotFoundError (HTTP 500) on direct attachment streaming

## Module
fs_attachment

## Bug
When the method _to_http_stream is called, for example when loading the /doc endpoint, there is a 500 error, and on the logs can be seen `Traceback: ir_attachment.py:921 os.stat(stream.path) → FileNotFoundError: /var/lib/odoo/filestore/mydatabase/s3://my_s3_bucket.`

This happens for example when S3 locations is being used, non detectable with regular filestore

## To Reproduce
**19.0**
1- Completely configure S3 filestore
2- Try to access /doc endpoint

Notice that this is how I found out, any other call to _to_http_stream should cause the same problem

Image

arielbarreiros96:
This fix is intended to fix #631 

Noticed the error when trying to load /doc endpoint

Changes are:
- super of _to_http_stream on ir.attachment
- small retouch on imports for preventing circular importing
- tests

### DIFF
--- a/fs_attachment/fs_stream.py
+++ b/fs_attachment/fs_stream.py
@@ -2,9 +2,12 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 from __future__ import annotations
 
+from typing import TYPE_CHECKING
+
 from odoo.http import STATIC_CACHE_LONG, Response, Stream, request
 
-from .models.ir_attachment import IrAttachment
+if TYPE_CHECKING:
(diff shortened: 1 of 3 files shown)


rrebollo (approved):
Code Review. LGTM!
Would you be so kind as to review mine (#634) in return?

PR #634, opened 2026-06-13
[17.0] [MIG] storage_backend_sftp

rrebollo:
We've been using #502 in production, but it never got merged. Let's try again.

#300 

@BinhexTeam T22451


arielbarreiros96 (changes requested):
The migration itself looks pretty straightforward, LGTM. The pre-commit is failing because of dependencies, once that is fixed i think it should be good to go 👌🏻

lmignon (approved):

rrebollo:
@lmignon I think you mean "merge"

PR #635, opened 2026-06-29
[18.0][IMP] fs_folder: Don't raise an error when the value is invalid or un…

lmignon:
…available

When the value of a fs_folder field is invalid or unavailable, it will now return a FsFolderInvalidValue or FsFolderUnavailableValue object instead of raising an error. This allows for more graceful handling of these cases in the code and more importantly, it will no longer break the form view when the value is invalid or unavailable.

### DIFF
--- a/fs_folder/fields.py
+++ b/fs_folder/fields.py
@@ -136,6 +136,84 @@ class FsContentValue:
         return not self.__eq__(other)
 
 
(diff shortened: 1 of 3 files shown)


lmignon:
ping @xavier-bouquiaux

lmignon:
@xavier-bouquiaux Can you provide some feedback regarding this PR?

Issue #639, opened 2026-07-06
[18.0][fs_attachment] On incomming helpdesk email [Errno 2] No such file or directory


CRogos:
## Module
fs_attachment

## Describe the bug
This problem occures with the Odoo Enterprise Helpdesk module, when a customer answers on a ticket. The eMail response cannot be added to the ticket thread, and the sender receives an error message via email.

@lmignon do you have an idea what is happening here?
I would expect some rename from  "azure-helpdesk://33/90/3390c0744f0da406a690ecdadff93121e4a9436e" to
"azure-helpdesk://33/90/logo_group_signatur_300dpi_377x134_c9a381b8-bb6d-43bf-bac2-7021ffd2ea88-605375-0.png", but not "azure-helpdesk/33/90/3390c0744f0da406a690ecdadff93121e4a9436e"?

## To Reproduce
**Affected versions**:
18.0

**Additional context**

```
2026-06-25 07:53:44,011 55954 ERROR c4a8-production-1830547 odoo.http: Exception during request handling. 
Traceback (most recent call last):
  File "/home/odoo/src/odoo/odoo/http.py", line 2577, in __call__
    response = request._serve_db()
               ^^^^^^^^^^^^^^^^^^^
  File "/home/odoo/src/odoo/odoo/http.py", line 2104, in _serve_db
    return self._transactioning(
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/odoo/src/odoo/odoo/http.py", line 2167, in _transactioning
    return service_model.retrying(func, env=self.env)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/odoo/src/odoo/odoo/service/model.py", line 157, in retrying
    result = func()
             ^^^^^^
  File "/home/odoo/src/odoo/odoo/http.py", line 2134, in _serve_ir_http
    response = self.dispatcher.dispatch(rule.endpoint, args)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/odoo/src/odoo/odoo/http.py", line 2295, in dispatch
    return self.request.registry['ir.http']._dispatch(endpoint)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/odoo/src/odoo/odoo/addons/base/models/ir_http.py", line 333, in _dispatch
[...]

PR #640, opened 2026-07-07
[18.0][IMP] fs_attachment: two attachments with same name and content

CRogos:
related to: https://github.com/OCA/storage/issues/639

### DIFF
--- a/fs_attachment/models/ir_attachment.py
+++ b/fs_attachment/models/ir_attachment.py
@@ -472,6 +472,7 @@ class IrAttachment(models.Model):
         Keeping the same meaning and mimetype is important to also ease to provide
         a meaningful and SEO friendly URL to the file in the filesystem storage.
         """
+        renamed_attachments = {}
         for attachment in self:
             if not self._is_file_from_a_storage(attachment.store_fname):
                 continue
@@ -485,7 +486,17 @@ class IrAttachment(models.Model):
             new_filename_with_path = os.path.join(
                 os.path.dirname(filename), new_filename
             )
-            fs.rename(filename, new_filename_with_path)
+
+            if filename in renamed_attachments:
+                if renamed_attachments[filename] == new_filename_with_path:
+                    # we already renamed this file, no need to rename it again
+                    continue
+                else:
+                    fs.copy(renamed_attachments[filename], new_filename_with_path)
+            else:
+                fs.rename(filename, new_filename_with_path)
+            renamed_attachments[filename] = new_filename_with_path
+
             attachment.fs_filename = new_filename
             # we need to update the store_fname with the new filename by
             # calling the write method of the field since the write method
--- a/fs_attachment/tests/test_fs_attachment.py
+++ b/fs_attachment/tests/test_fs_attachment.py
@@ -431,6 +431,56 @@ class TestFSAttachment(TestFSAttachmentCommon):
             }
         )
 
+    def test_create_two_attachments(self):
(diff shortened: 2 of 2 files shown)


fs_attachment/tests/test_fs_attachment.py
```diff
@@ -431,6 +431,24 @@ def test_create_attachments_basic_user(self):
...
+        )
+
+    def test_create_two_attachments_same_name(self):
+        self.temp_backend.use_as_default_for_attachments = True
+        self.ir_attachment_model.create(
+            [
+                {"name": "test.txt", "raw": b"content"},
+                {"name": "test.txt", "raw": b"content"},
+            ]
+        )
```
lmignon: Can you check you've 2 attachment records with 2 different files.  If I update one file, the other one should not be changed.

fs_attachment/tests/test_fs_attachment.py
```diff
@@ -431,6 +431,24 @@ def test_create_attachments_basic_user(self):
...
         )
 
+    def test_create_two_attachments(self):
+        self.temp_backend.use_as_default_for_attachments = True
+        self.ir_attachment_model.create(
+            [
+                {"name": "test2.txt", "raw": b"content"},
+                {"name": "test.txt", "raw": b"content"},
+            ]
+        )
```
lmignon: OK the command create works  but can you check the result. Otherwise your tests is partially useless....

fs_attachment/models/ir_attachment.py
```diff
@@ -485,7 +486,17 @@ def _enforce_meaningful_storage_filename(self) -> None:
             new_filename_with_path = os.path.join(
                 os.path.dirname(filename), new_filename
             )
-            fs.rename(filename, new_filename_with_path)
```
lmignon, line 488: With your change, attachments will share the same file. What If we try to later create a new attachment with the same name and content....

lmignon (changes requested):
Thank you for pointing out this use case, which I hadn't anticipated. 
There are several factors to consider:
* The same file with the same name can be created in different calls to “create.” (This case is not handled.)
* We need to ensure that deleting an attachment does not delete the file referenced by other attachments (this should not be the case with the garbage collector mechanism)
* We need to ensure that if I write directly to the file (my_file.open()), only the attachment I wrote to is modified. (This should be the case)

fs_attachment/models/ir_attachment.py
CRogos: I agree this might not be the perfect solution, but are you sure? The source file is only available once, and is renamed with the first file, but the target is copied to the second destination on the second file.

fs_attachment/tests/test_fs_attachment.py
CRogos: The previous result was an exception, but I agree we should so some check.

fs_attachment/tests/test_fs_attachment.py
CRogos: Is this change OK?

CRogos (commented):
@lmignon do you have another solution in mind?

fs_attachment/models/ir_attachment.py
lmignon: I've to dig into the code ...

lmignon (approved):
Thank you for the fix, @CRogos. After digging into the code, I fully understood the issue, and your fix is indeed the right one.

lmignon:
Thank you for your work @CRogos :smirk:

PR #645, opened 2026-07-23
[16.0][MIG] storage_backend_ftp: Migration to 16.0

eldhose-beo:
This PR migrates `storage_backend_ftp` to 16.0, rebased on the real upstream
15.0 history.

Ref: ADOO-2794


lmignon (approved):
You should take a look to fs_storage.... ftp is supported by default. But you may be need this one to ease your migration from a previous version

eldhose-beo:
> You should take a look to fs_storage.... ftp is supported by default. But you may be need this one to ease your migration from a previous version

Thanks for the review and the pointer to fs_storage, @lmignon!

You're right that fs_storage covers FTP natively via fsspec for new setups. In our case though, we have several production systems still running storage_backend_ftp from earlier Odoo versions, and this module is what lets us carry that configuration forward cleanly while we migrate them to 16.0 — hence the short-term need. A move to fs_storage for FTP is definitely the right long-term direction, and we'll track that as a separate follow-up once the 16.0 migration is settled.

One more thing before this goes further: while double-checking the branch internally we found the history was based on the wrong upstream branch (13.0 instead of 15.0, which is where this module was last maintained upstream), which dragged in some regressions — dropped tests, gutted README, and a real bug in the plain-FTP path (ftplib.FTP(timeout=False) puts the socket in non-blocking mode, so it can't connect at all). I'm pushing a corrected version rebased on the right history shortly. Would appreciate a fresh look once that's up rather than relying on the current diff — sorry for the extra round trip.

lmignon (approved):

lmignon (changes requested):
@eldhose-beo Can you fix the test? Everything looks good expect a little change required regarding the place from where `mock`must be now imported  to do to make the tests green .

eldhose-beo:
> @eldhose-beo Can you fix the test? Everything looks good expect a little change required regarding the place from where `mock`must be now imported to do to make the tests green .

@lmignon Fixed — the mock import issue is resolved in [4ff75b0](https://github.com/AmetrasIntelligence/storage/commit/4ff75b0) (switched to from unittest import mock, since the standalone package was dropped and unittest.mock covers the same API). Both test jobs are green now. Re-requesting your review — thanks for catching that.

lmignon (approved):

QuatschSalat (approved):

PR #650, opened 2026-08-03
[IMP] fs_image_*: Defines images in all expected sizes into mixin

lmignon:
By default odoo defines into its own 'image.mixin' model, computed images in 1024, 512, 256 and 128 pixels from original images. Align our specific mixin to follow the same convention


lmignon:
Closing... not the right approach...

PR #651, opened 2026-08-04
[ADD] attachment_pdfa

baimont:
Here's a proposal for a new module that will allow to convert PDF attachments to PDF/A.
It does nothing on itself.
Developer has to implement which models must opt-in by inheriting `attachment.pdfa.mixin` and override `_attachment_must_be_pdfa()`.


baimont:
@sbidoul and @AnizR please take a look?

attachment_pdfa/models/ir_attachment.py
```diff
@@ -0,0 +1,215 @@
...
+        except Exception as e:
+            msg = f"Odoo PDF/A conversion failed: {str(e)}"
+            _logger.exception(msg)
+            return raw_content, False, msg
+
+    def _should_convert_to_pdfa(self, vals):
+        """Check if attachment creation dictionary qualifies for conversion."""
+        mimetype = vals.get("mimetype", "")
+        name = vals.get("name", "")
+        if not (mimetype == "application/pdf" or name.lower().endswith(".pdf")):
```
AnizR:
```suggestion
        if not (mimetype == "application/pdf" or not name.lower().endswith(".pdf")):
```

attachment_pdfa/models/ir_attachment.py
```diff
@@ -0,0 +1,215 @@
...
+                        new_pdf_stream.getvalue(),
+                        True,
+                        "Odoo conversion successful.",
+                    )
+        except Exception as e:
+            msg = f"Odoo PDF/A conversion failed: {str(e)}"
+            _logger.exception(msg)
+            return raw_content, False, msg
+
+    def _should_convert_to_pdfa(self, vals):
```
AnizR: `@api.model`

attachment_pdfa/models/ir_attachment.py
```diff
@@ -0,0 +1,215 @@
...
+            f"<pre style='{style}'>{html_escape(log_msg)}</pre>"
+        )
+        target_record.message_post(
+            body=Markup(formatted_body),
+            message_type="notification",
+            subtype_xmlid="mail.mt_note",
+        )
+
+    @api.model_create_multi
+    def create(self, vals_list):
```
AnizR: What about the write function?

Odoo has some workflows where they create the attachment then do a write to link to a record. IMO, in this case, the conversion should also be applied.

attachment_pdfa/models/ir_attachment.py
```diff
@@ -0,0 +1,215 @@
...
+            msg = "Ghostscript ICC Profile or PDF/A definition file not found."
+            _logger.error(msg)
+            return raw_content, False, msg
+        ps_content = icc_profile_ps.read_text(encoding="utf-8", errors="replace")
+        ps_content = re.sub(
+            r"/ICCProfile \((.+)\)",
+            f"/ICCProfile ({rgb_profile})",
+            ps_content,
+        )
+        try:
```
AnizR: Is it really necessary modify the `icc_profile_ps` each time?
If not, I am wondering if the 'temp_ps' below is still needed

AnizR (commented):
Good contribution 🚀

attachment_pdfa/models/ir_attachment.py
```diff
@@ -0,0 +1,215 @@
...
+
+    def _gs_convert_pdf_to_pdfa(self, raw_content):
+        """Convert PDF raw bytes using Ghostscript."""
+        data_dir = Path(__file__).parent.parent / "data"
+        icc_profile_ps = data_dir / "pdfa_def.ps"
+        rgb_profile = data_dir / "iccprofiles_default_rgb.icc"
+        if not icc_profile_ps.is_file() or not rgb_profile.is_file():
+            msg = "Ghostscript ICC Profile or PDF/A definition file not found."
+            _logger.error(msg)
+            return raw_content, False, msg
```
lmignon: These paths should be initialized and checked when the module is imported. (at the top of the file).

attachment_pdfa/models/ir_attachment.py
```diff
@@ -0,0 +1,215 @@
...
+        mimetype = vals.get("mimetype", "")
+        name = vals.get("name", "")
+        if not (mimetype == "application/pdf" or name.lower().endswith(".pdf")):
+            return False
+        res_model = vals.get("res_model") or self.env.context.get("default_res_model")
+        res_id = vals.get("res_id") or self.env.context.get("default_res_id")
+        if not (res_model and res_id):
+            return False
+        model_obj = self.env.get(res_model)
+        if model_obj is None or not hasattr(model_obj, "_attachment_must_be_pdfa"):
```
lmignon: This check should be put into a dedicated method

attachment_pdfa/models/ir_attachment.py
```diff
@@ -0,0 +1,314 @@
...
+        temp_in = tempfile.NamedTemporaryFile(suffix=".pdf", delete=False)
+        temp_out = tempfile.NamedTemporaryFile(suffix=".pdf", delete=False)
+        try:
+            temp_in.write(raw_content)
+            temp_in.close()
+            temp_out.close()
+            args = [
+                "gs",
+                f"--permit-file-read={RGB_PROFILE}",
+                "-dPDFACompatibilityPolicy=1",
```
AnizR:
```suggestion
                "-dPDFACompatibilityPolicy=2",
```

Seems to low, if an error happens during conversion, it should create an error and the conversion should be done using the fallback function.
From GS doc: https://ghostscript.readthedocs.io/en/latest/VectorDevices.html#creating-a-pdf-a-document

PR #653, opened 2026-08-19
[16.0][FIX] fs_attachment,fs_attachment_s3,fs_storage,fs_storage_backup: fix server_environment split migration and add install/uninstall hooks

yankinmax:
The PR was merged:
- [ ] https://github.com/OCA/storage/pull/627

This PR fixes the broken migration scripts, which set `ir_module_module::state` directly via SQL instead of calling `button_install`, so the new `*_environment` modules never actually got installed on upgrade.
Scripts are moved to the `upgrades/` folder (same version, no version bump) and fixed to call `button_install`.

Also adds `uninstall_hook` to each `*_environment module`, using `restore_env_managed_columns` from `server_environment` to restore the DB columns dropped by `server.env.mixin` when the module is uninstalled.

Depends on:
- [ ] https://github.com/OCA/server-env/pull/276


ivantodorovich (approved):

lmignon:
I haven't had any upgrade issues with the current implementation.... which doesn't mean the fix doesn't improve the robustness of the approach.

yankinmax:
`image_tag` split is handled separately, because it doesn't introduce any `server_env_fields` override, so only need the upgrade script:
- [ ] https://github.com/OCA/storage/pull/575

lmignon (approved):

yankinmax:
To be able to fix the issue we need to merge the dependencies, because they're conflicting now:
```
ERROR: Cannot install odoo-addon-server-environment 16.0.1.1.2.6 (from git+https://github.com/OCA/server-env.git@refs/pull/276/head#subdirectory=setup/server_environment) and odoo-addon-server-environment 16.0.1.1.2.6 (from git+https://github.com/OCA/server-env.git@refs/pull/287/head#subdirectory=setup/server_environment) because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested odoo-addon-server-environment 16.0.1.1.2.6 (from git+https://github.com/OCA/server-env.git@refs/pull/276/head#subdirectory=setup/server_environment)
    The user requested odoo-addon-server-environment 16.0.1.1.2.6 (from git+https://github.com/OCA/server-env.git@refs/pull/287/head#subdirectory=setup/server_environment)
```

**UPDATE**
This is solved by the combining dependencies into 1 PR:
- [ ] https://github.com/OCA/server-env/pull/276

PR #654, opened 2026-08-19
[17.0] [IMP] storage: Remove dependency on server_environment

yankinmax:
This is a forward-port of:
- [ ] https://github.com/OCA/storage/pull/627

With the applied fixes introduced in:
- [ ] https://github.com/OCA/storage/pull/653
- [ ] https://github.com/OCA/server-env/pull/286


yankinmax:
Hello @etobella @lmignon @ivantodorovich 
In this PR I've cherry-picked the commits from merged 16.0 PR.
I squashed the commits that are related to one module split.
I've added new commits for `storage_backend*` modules.
The upgrade scripts and uninstall hooks are added for each module.
`image_tag` split is handled separately, because it doesn't introduce any `server_env_fields` override, so only need the upgrade script:
- [ ] https://github.com/OCA/storage/pull/580

lmignon (approved):
Thank you for the FWP

ivantodorovich (approved):
Thanks!

fs_storage_environment/__manifest__.py
```diff
@@ -0,0 +1,18 @@
...
+    "category": "FS Storage",
+    "website": "https://github.com/OCA/storage",
+    "author": " ACSONE SA/NV, Dixmit, Odoo Community Association (OCA)",
+    "license": "LGPL-3",
+    "development_status": "Beta",
+    "installable": True,
+    "depends": ["fs_storage", "server_environment"],
+    "post_init_hook": "post_init_hook",
+    "uninstall_hook": "uninstall_hook",
+    "auto_install": True,
```
simahawk: Hmm are we sure we want to make these modules auto-install? 
That means that if I still have server env installed for a different app (eg: mail) this module will be installed as well w/o any reason.
Since the migration is taking care of installing the modules I would say this has to change.
Am I missing any particular reason to keep auto-install?

simahawk (changes requested):

fs_storage_environment/__manifest__.py
yankinmax: The reason was to keep it as much as possible similar to what it was before, so making them auto install would prove the goal, but there is no specific reason.
WDYT @ivantodorovich @lmignon @etobella ?

lmignon: For the next version of Odoo, we should remove the `autoinstall`. However, since this feature was enabled by default in existing versions, I think we should preserve the current default behaviour for backward compatibility.

BTW, I recently discovered the root cause of a strange behaviour with `server.env.mixin`, where some fields can no longer be edited manually. see https://github.com/OCA/server-env/issues/289

yankinmax: Finally, I've decided to remove `auto_install` and tend to agree with @simahawk :
All five upgrade scripts explicitly call `button_install`.
Removing `auto_install`:
1. Preserves environment managed fields and previous behavior during upgrades.
2. Prevents fresh installations from unexpectedly installing a `storage_environment_*` module just because `server_environment` is used a dependency by another module.
3. The `auto_install` has never landed to recent versions actually. It was an idea to add it, but it seems there is no practical need.

simahawk (approved):

simahawk:
The dependency is getting merged :wink:

yankinmax:
> The dependency is getting merged 😉

I'll update this PR without `auto_install` just after

simahawk:
@yankinmax ready for merge w/ nobump, right?

yankinmax:
> @yankinmax ready for merge w/ nobump, right?

Yep, the version is bumped manually accordingly with upgrade script

yankinmax:
@simahawk can you pls help to investigate the merge issue? Can I do anything?

etobella:
@yankinmax there is a limitation of the number of new packages that we can push in a day from the community. Let's check if it passes now, otherwise, we have a second option to ask for some help to make it happen.

yankinmax:
Thank you @etobella for the explanation, good to know)!

etobella:
Ok, we will need to ask for help then :cry: 

@sbidoul

sbidoul:
name reserved on PyPI

etobella:
:heart: thanks @sbidoul 

/ocabot merge nobump

PR #662, opened 2026-08-27
[IMP] Add FS Storage Dynamic Routing Rules

baimont:


baimont:
take a look @sbidoul and @AnizR

AnizR:
> take a look @sbidoul and @AnizR

Can you give some context?

lmignon (changes requested):
@baimont Could you add some user-oriented documentation explaining this new feature in the usage section?

I think this is particularly important here, as without this explanation it is quite difficult to understand what the code is doing, and someone discovering the module would otherwise have no way of knowing that this feature exists.

baimont:
> @baimont Could you add some user-oriented documentation explaining this new feature in the usage section?
> 
> I think this is particularly important here, as without this explanation it is quite difficult to understand what the code is doing, and someone discovering the module would otherwise have no way of knowing that this feature exists.

hey @lmignon, done

fs_attachment/readme/USAGE.md
```diff
@@ -89,6 +89,45 @@ to different resource fields/models. You can configure it either on the
...
+
+1. The first matching dynamic `fs.storage.rule`, if any.
+2. The static `model_ids`/`field_ids` mapping (or `model_xmlids`/`field_xmlids`
+   from a server environment file).
+3. The storage configured as the default for attachments.
+
+Note
+
+A rule's domain is only evaluated when the attachment is created or when its
+content is rewritten (e.g. via `open(..., new_version=True)`). It is **not**
```
lmignon: `new_version=True` is the default. IMO, it's important to be precise here so that users don't think that a regular call to open() without this parameter will behave differently.

fs_storage/models/fs_storage.py
```diff
@@ -411,6 +412,36 @@ def get_fs_by_code(self, code, force_no_cache=False):
...
+    @tools.ormcache("model_name")
+    def _get_storage_rule_ids_for_model(self, model_name):
+        """Static/cacheable part: which rules exist for this model."""
+        return tuple(
+            self.env["fs.storage.rule"]
+            .sudo()
+            .search([("model_id.model", "=", model_name)])
+            .ids
+        )
+
```
lmignon, line 436: If you create a cache of rule ids for a given model_name, you should query the database only once for all the model_name and keep a dictionary of ids by model_name into the cache

fs_storage/models/fs_storage.py
```diff
@@ -411,6 +412,36 @@ def get_fs_by_code(self, code, force_no_cache=False):
...
+        """
+        if model_name and res_id:
+            rule_ids = self._get_storage_rule_ids_for_model(model_name)
+            if rule_ids:
+                record = self.env[model_name].sudo().browse(res_id).exists()
+                if record:
+                    for rule in self.env["fs.storage.rule"].sudo().browse(rule_ids):
+                        if rule.field_id and rule.field_id.name != field_name:
+                            continue
+                        if record.filtered_domain(safe_eval(rule.domain or "[]")):
```
lmignon: The cache of ids for a model_name is useless if you need to browse the records to get the field name and the domain. Why not a cache of [(field_name, storage_code, domain) by model_name?

lmignon (changes requested):
Some little comments regarding the way you design your cache....

simahawk (commented):
@lmignon am I wrong to say that this does not sound as a core feature and can stay in its own module?

lmignon:
> @lmignon am I wrong to say that this does not sound as a core feature and can stay in its own module?

You're right. I've the feeling it could be developed as an extension... @baimont What do you thing about?

baimont:
> > @lmignon am I wrong to say that this does not sound as a core feature and can stay in its own module?
> 
> You're right. I've the feeling it could be developed as an extension... @baimont What do you thing about?

I hesitated but IMO this is an improvement of the model_ids and field_ids fields on fs.storage that already allow to route to different storages. That's why I went for this. I can change it.

lmignon:
> > > @lmignon am I wrong to say that this does not sound as a core feature and can stay in its own module?
> > 
> > 
> > You're right. I've the feeling it could be developed as an extension... @baimont What do you thing about?
> 
> I hesitated but IMO this is an improvement of the model_ids and field_ids fields on fs.storage that already allow to route to different storages. That's why I went for this. I can change it.

The discussion is  open :smirk:  The main goal is to avoid increasing the module's complexity by adding optional features that are not required for common use cases. Maybe we could also rethink the structure of this addon for Odoo 20 and split it into smaller modules.

AnizR (commented):
IMO, the feature can be in a separate module. (And so, I'll wait before approving the PR)

Otherwise, the code seems pretty clean to me, good job 🚀

